PR #28, opened 2026-04-12
feat: add Nix flake for package distribution

PJalv:
## Summary
- Adds Nix flake for installing `mex` without requiring `npx`
- Enables direct integration into NixOS configurations

## Features
- `buildNpmPackage` derivation for the CLI
- NixOS module for easy system integration (`programs.mex.enable = true`)
- Dev shell with Node.js 22
- Overlay for use in other flakes

### NixOS Configuration
```nix
{
  inputs.mex.url = "github:PJalv/mex";
  
  outputs = { nixpkgs, mex, ... }: {
    nixosConfigurations.your-system = nixpkgs.lib.nixosSystem {
      modules = [
        mex.nixosModules.default
        { programs.mex.enable = true; }
      ];
    };
  };
}
```

### Ad-hoc Usage
```bash
nix run github:PJalv/mex
```

### Dev Shell
```bash
nix develop github:PJalv/mex
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,8 @@
 node_modules/
 dist/
 
+# Nix
+result
+
 # OS
 Thumbs.db
--- a/flake.lock
+++ b/flake.lock
@@ -1,0 +1,27 @@
+{
+  "nodes": {
+    "nixpkgs": {
+      "locked": {
+        "lastModified": 1775710090,
+        "narHash": "sha256-ar3rofg+awPB8QXDaFJhJ2jJhu+KqN/PRCXeyuXR76E=",
+        "owner": "NixOS",
+        "repo": "nixpkgs",
+        "rev": "4c1018dae018162ec878d42fec712642d214fdfa",
+        "type": "github"
+      },
+      "original": {
+        "owner": "NixOS",
+        "ref": "nixos-unstable",
+        "repo": "nixpkgs",
+        "type": "github"
+      }
+    },
+    "root": {
+      "inputs": {
+        "nixpkgs": "nixpkgs"
+      }
+    }
+  },
+  "root": "root",
+  "version": 7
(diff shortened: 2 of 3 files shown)


theDakshJaitly (commented):
Hey @PJalv, thanks for this — nice contribution for Nix users! A few things to address before merging:

### Version mismatch

`flake.nix:22` hardcodes `version = "0.4.0"` but `package.json` is currently at `0.3.2`. This should match. Ideally it would be derived from `package.json` so it doesn't drift on every release, but at minimum it needs to be `0.3.2` right now.

### `npmDepsHash` maintenance

`flake.nix:25` — this hash will break any time `package-lock.json` changes. Could you add a comment above it noting that? Something like:

```nix
# Update this hash when package-lock.json changes:
# nix build 2>&1 | grep 'got:' | awk '{print $2}'
npmDepsHash = "sha256-...";
```

Just so future contributors know what to do when the build fails.

### Add `result` to `.gitignore`

`nix build` creates a `result` symlink in the project root. Should add that to `.gitignore` so it doesn't get accidentally committed.

### Minor note

The `postInstall` path (`$out/lib/node_modules/promexeus/templates/`) uses the npm package name `promexeus` — that's correct today, just flagging it as something that would need updating if the package name ever changes.

---

Everything else looks solid — platform coverage, the NixOS module, overlay, and dev shell all follow standard patterns. Nice work.

PJalv:
Thanks for checking it out!
I will definitely be making sure the flake correctly reflects the current state of the package; 
Already trying it out with my current AI workflow!

theDakshJaitly (approved):
All review feedback addressed — version aligned to 0.3.2, npmDepsHash comment added, result in .gitignore. Looks good, thanks @PJalv!

PR #34, opened 2026-04-19
fix(staleness): collapse day + commit thresholds into a single issue

mvanhorn:
## Summary

`checkStaleness` returned two separate `STALE_FILE` issues when a file tripped both the day threshold (>=90d) and the commit threshold (>=200 commits). `computeScore` deducts per-issue, so the same underlying staleness cost the score 20 points instead of 10 (a file can drop from 100 to 0 off a single stale scaffold if other issues are also present).

This collapses the two signals into a single compound `STALE_FILE` issue using:
- the higher of the two severities (error wins over warning), and
- a combined message that preserves both data points (`"...100 days...; ...250 commits..."`)

`computeScore` is unchanged -- the fix keeps the scoring layer pure and localized to the checker.

## Test plan

- [x] 6 new unit tests in `test/checkers.test.ts` covering:
  - clean file (no issues)
  - only the day threshold tripped
  - both thresholds tripped -> one combined issue
  - warning + error thresholds -> error wins
  - both warnings -> stays warning
  - `null` git history -> no issues
- [x] Existing suite: 100/100 tests passing
- [x] `npm run build` clean
- [x] `npx tsc --noEmit` clean

Closes #31

Issue #31, opened 2026-04-15
checkStaleness returns two independent STALE_FILE errors for a single stale file, causing double score penalty


shridipavansaber:
## Description

When a file trips both the day threshold (≥90 days) and the commit threshold (≥200 commits), \`checkStaleness\` returns two separate \`STALE_FILE\` issues for the same underlying condition. Because \`computeScore\` deducts 10 points per error, a single stale file can silently subtract 20 points from the drift score instead of 10.

## Reproduction

\`\`\`typescript
// checkStaleness("file.md", "source.md", cwd) with days=100, commits=250
// returns:
[
  { code: "STALE_FILE", severity: "error", message: "File hasn't been updated in 100 days..." },
  { code: "STALE_FILE", severity: "error", message: "250 commits since file was last updated..." }
]

// computeScore([issue1, issue2]) deducts 10 + 10 = 20
\`\`\`

A file can score as low as 0 from a single stale file if other issues are also present, while a project with the same file but only one threshold exceeded scores 10 points higher — even though the staleness is qualitatively identical.

## Expected behaviour

Either:
- \`checkStaleness\` returns a single issue using the higher severity of the two thresholds, or
- \`computeScore\` is made aware that two \`STALE_FILE\` issues on the same file should be treated as one compound condition

## Notes

Found while running [tailtest](https://github.com/avansaber/tailtest) on this codebase. The generated test suite documents the current behaviour: \`expect(issues).toHaveLength(2)\` when both thresholds are exceeded.

theDakshJaitly:
Thank you for bringing this up, will be fixed asap.

PR #35, opened 2026-04-20
feat(drift): make staleness thresholds configurable (#6)

### LINKED ISSUE #6: Make staleness checker thresholds configurable
## Problem

The staleness checker (`src/drift/checkers/staleness.ts`) uses hardcoded thresholds: 30 days and 50 commits. Different projects have different update cadences — a fast-moving startup might want 7 days, a stable open-source project might be fine with 90.

## What to do

Make the thresholds configurable. Options:

- CLI flags: `mex check --stale-days 14 --stale-commits 20`
- Config file (e.g., `.mex/config.json` or a frontmatter field)
- Both (CLI overrides config)

## Where to look

- `src/drift/checkers/staleness.ts` — current hardcoded values
- `src/cli.ts` — if adding CLI flags
- `src/types.ts` — `MexConfig` type if adding config file support

## Acceptance criteria

- Users can set custom staleness thresholds
- Defaults remain 30 days / 50 commits when not configured
- Add tests for custom thresholds

mvanhorn:
Closes #6.

## Summary

Thresholds used by the staleness checker (`warnDays` / `errorDays` / `warnCommits` / `errorCommits`) were hardcoded at 30 / 90 / 50 / 200. Fast-moving projects want a tighter bar; stable projects want a looser one. They're now configurable via three precedence layers (highest first):

1. **CLI flags** on `mex check`:
   - `--stale-warn-days <n>` / `--stale-error-days <n>`
   - `--stale-warn-commits <n>` / `--stale-error-commits <n>`
2. **Config file** (`.mex/config.json`) under a new `staleness` key:
   ```json
   { "aiTools": ["claude"], "staleness": { "warnDays": 14, "errorDays": 30 } }
   ```
   Partial overrides are allowed — any missing field falls back to the default.
3. **Defaults** — exported as `DEFAULT_STALENESS_THRESHOLDS` (30 / 90 / 50 / 200), unchanged from the previous constants.

## Why four flags, not two

The issue's example (`--stale-days 14 --stale-commits 20`) suggests two knobs, but the checker has four thresholds (warn + error for each of days and commits) and the relationship between them isn't a clean multiplier (90/30=3, 200/50=4). Exposing all four keeps the CLI honest and lets users tighten just the warn bar without moving the error bar. Happy to collapse to two if you'd prefer — the shape of the data flow is unchanged either way.

## Internals

- `StalenessThresholds` added to `src/types.ts`.
- `MexConfig.stalenessThresholds` is optional; `findConfig` loads it from `.mex/config.json` via a new `loadStalenessThresholds` helper that validates each field as a non-negative number and falls back to defaults for any missing field.
- `checkStaleness(...)` takes thresholds as an optional fourth arg, default `DEFAULT_STALENESS_THRESHOLDS`.
- `runDriftCheck` threads `config.stalenessThresholds` through to the checker.
- `mex check` validates flag values via commander's `InvalidArgumentError` so `--stale-warn-days foo` fails fast with a clear message.

## Tests

`test/staleness.test.ts` — 8 new cases, all passing:

- `DEFAULT_STALENESS_THRESHOLDS` exports the documented defaults (30/90/50/200).
- Silent when the file is fresh.
- Warnings at the default warn bar (31d / 60c).
- Errors at the default error bar (120d / 300c), with thresholds printed in the message.
- Custom `warnDays=14` triggers at 15d with a `threshold: 14d` message.
- Custom `errorDays=30` triggers at 45d with a `threshold: 30d` message.
- Custom commit-only thresholds isolate from day thresholds.
- Silent when custom thresholds are raised above the reality.

All 102 tests pass. `npm run typecheck` and `npm run build` also succeed.

## Not in scope

- Exposing thresholds to `mex setup` / `mex sync`. If you want `mex setup` to prompt for staleness preferences, that's a separate UX change.
- Per-file overrides. Current config is repo-wide.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,7 +1,15 @@
 import chalk from "chalk";
-import { Command } from "commander";
+import { Command, InvalidArgumentError } from "commander";
 import { findConfig } from "./config.js";
 import { reportConsole, reportQuiet, reportJSON, reportVerbose } from "./reporter.js";
+
+function parseIntArg(raw: string): number {
+  const n = Number.parseInt(raw, 10);
+  if (!Number.isFinite(n) || n < 0) {
+    throw new InvalidArgumentError(`Expected a non-negative integer, got "${raw}".`);
+  }
+  return n;
+}
 
 const program = new Command();
 
@@ -33,11 +41,27 @@ program
   .option("--quiet", "Single-line summary only")
   .option("--fix", "Run sync to fix any issues found")
   .option("--verbose", "Show detailed diagnostic output")
+  .option("--stale-warn-days <n>", "Warn when a file hasn't changed in N days (default 30)", parseIntArg)
+  .option("--stale-error-days <n>", "Error when a file hasn't changed in N days (default 90)", parseIntArg)
+  .option("--stale-warn-commits <n>", "Warn when a file has N commits since its last change (default 50)", parseIntArg)
+  .option("--stale-error-commits <n>", "Error when a file has N commits since its last change (default 200)", parseIntArg)
   .action(async (opts) => {
     try {
       const config = findConfig();
       const { runDriftCheck } = await import("./drift/index.js");
-      const report = await runDriftCheck(config, { verbose: opts.verbose });
+      const { DEFAULT_STALENESS_THRESHOLDS } = await import("./drift/checkers/staleness.js");
+
+      const stalenessThresholds = {
+        warnDays: opts.staleWarnDays ?? config.stalenessThresholds?.warnDays ?? DEFAULT_STALENESS_THRESHOLDS.warnDays,
+        errorDays: opts.staleErrorDays ?? config.stalenessThresholds?.errorDays ?? DEFAULT_STALENESS_THRESHOLDS.errorDays,
+        warnCommits: opts.staleWarnCommits ?? config.stalenessThresholds?.warnCommits ?? DEFAULT_STALENESS_THRESHOLDS.warnCommits,
+        errorCommits: opts.staleErrorCommits ?? config.stalenessThresholds?.errorCommits ?? DEFAULT_STALENESS_THRESHOLDS.errorCommits,
+      };
+
+      const report = await runDriftCheck(
+        { ...config, stalenessThresholds },
+        { verbose: opts.verbose },
+      );
 
       if (opts.json) {
         reportJSON(report, { verbose: opts.verbose });
--- a/src/config.ts
+++ b/src/config.ts
@@ -1,6 +1,7 @@
 import { existsSync, readFileSync, writeFileSync, mkdirSync } from "node:fs";
 import { resolve, dirname } from "node:path";
 import type { MexConfig, AiTool, StalenessThresholds } from "./types.js";
+import { DEFAULT_STALENESS_THRESHOLDS } from "./drift/checkers/staleness.js";
 
 /**
  * Walk up from startDir looking for .git to find project root,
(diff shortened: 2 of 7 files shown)


theDakshJaitly (changes requested):
Thanks for this @mvanhorn — overall really nice shape. A few things before I can merge.

**The main blocker: this conflicts with your own PR #34.**

#34 refactored `checkStaleness` to collapse the day and commit signals into a single compound issue (fixing the double-score-penalty bug in #31). This PR is built on top of the old four-branch if/else and just parameterizes the constants. Two concrete consequences after #34 merges:

1. Merge conflict in `src/drift/checkers/staleness.ts` — both PRs rewrite the same block.
2. Two tests in `test/staleness.test.ts` will fail because they assume the pre-#34 behavior:
   - `"emits warnings at the default warn thresholds"` — asserts `toHaveLength(2)` for days=31 + commits=60
   - `"emits errors at the default error thresholds"` — asserts `toHaveLength(2)` for days=120 + commits=300

   Both become `toHaveLength(1)` with a compound message under #34's model.

I'd like to merge #34 first, then have you rebase this on top and update those two tests to match the collapsed-issue behavior.

**Smaller things while you're in there (not blocking):**

- In `src/config.ts`, `loadStalenessThresholds` hardcodes the fallbacks `30/90/50/200` inline. Could you import `DEFAULT_STALENESS_THRESHOLDS` from the staleness checker and spread that instead? Keeps the defaults defined in exactly one place.
- No invariant check that `errorDays >= warnDays` (same for commits). A misconfigured `warnDays: 90, errorDays: 30` silently makes the warn path unreachable. Not a regression vs. the hardcoded values, but since the config is now user-facing, a small validation in `loadStalenessThresholds` (warn once and fall back to defaults) would be friendlier. Happy to leave this for a follow-up if you'd rather keep the PR tight.

**Things I liked:**

- Going with 4 flags instead of the 2 from the issue. The warn/error ratios aren't a clean multiplier (3x for days, 4x for commits) so collapsing would've been lossy. Good call, and the PR body explains the reasoning well.
- Precedence chain (CLI > config > defaults) reads cleanly.
- Non-negative integer validation via `InvalidArgumentError` → users get a clear error on bad input.
- Tests use module-level `vi.mock` cleanly, same pattern as #34.

Let me know when you've rebased.

mvanhorn:
Rebased and pushed, @theDakshJaitly:

- 3da554e: rebased onto upstream/main on top of #34. The two staleness tests that asserted `toHaveLength(2)` now assert `toHaveLength(1)` with a combined day+commit message, matching the compound-issue shape from #34.
- 06cb4c9: applied both non-blocking items. `loadStalenessThresholds` now imports `DEFAULT_STALENESS_THRESHOLDS` from the staleness checker and spreads from it, so the defaults live in one place. Added an invariant guard: if `errorDays < warnDays` or `errorCommits < warnCommits`, the loader warns once and falls back to defaults rather than silently disabling the warn path. 4 new tests in `test/config.test.ts` cover the full-load, partial-fill, and both invariant-violation cases.

112/112 tests pass locally, `tsc --noEmit` clean. Ready for another look when you have time.

theDakshJaitly (approved):
LGTM. Rebase onto #34 is clean, tests match the compound-issue shape, and both non-blocking items (shared DEFAULT_STALENESS_THRESHOLDS + warn/error invariant guard) are in with coverage. CI green on 18/20/22.

mvanhorn:
Thanks for landing the configurable thresholds.

PR #44, opened 2026-05-14
export field configured by  exposing a stable public API surface for mex-agent

Yashasvi2229:
What

  Expose a stable public API surface for mex-agent so embedders can consume it in-process as an npm   
  dependency, instead of shelling out to the mex CLI.

  The change is purely additive:

  - New src/index.ts — curated public entry re-exporting drift, events, heartbeat, frontmatter, config
   helpers, and shared types.
  - tsup.config.ts split into two builds — CLI (dist/cli.js, shebang banner) and library
  (dist/index.js + dist/index.d.ts).
  - package.json gains main, types, and exports fields. bin unchanged.
  - createConfig({ projectRoot, scaffoldRoot, aiTools? }) factory for callers that already know where 
  the project + scaffold live (bypasses findConfig's .mex/ discovery).
  - runDriftCheck and checkHeartbeat/runHeartbeat accept an optional scaffoldPatterns override, with  
  DEFAULT_SCAFFOLD_PATTERNS and DEFAULT_HEARTBEAT_PATTERNS exported so callers can extend rather than 
  replace.
  - EVENT_KINDS promoted from a private Set to a public runtime constant (EventKind type now derives  
  from it).
  - New test/public-api.test.ts — 15-case smoke test that imports only from the public surface and    
  asserts the contract (function shapes, runtime constants, round-trip behaviour). CI-enforced so the 
  contract can't rot silently.

  Why

  mex-agent currently ships only a CLI binary. External tools that want to reuse its drift detector,  
  event log, heartbeat checker, or frontmatter parser have two bad options: shell out to mex (slow, no
   types, fragile across versions) or vendor the source (creates a maintenance fork). Adding a typed  
  public entry lets them depend on mex-agent directly via npm install and import functions in-process.

  T-Rex (a team-mode layer being built on top of mex) is the first such consumer and motivated the    
  specific export list, but nothing here is T-Rex-specific.

  Type of change

  - New feature
  - CI/Tooling

  How to test

  1. npm run typecheck — should pass clean.
  2. npm test — new test/public-api.test.ts should pass all 15 cases. Pre-existing Windows
  path-separator failures in test/heartbeat.test.ts:38 and test/scanner.test.ts:56 are unrelated to   
  this PR and exist on main.
  3. npm run build — verify dist/ contains cli.js, index.js, and index.d.ts.
  4. From any external project: npm install <path-to-this-checkout> then import { runDriftCheck,      
  createConfig, EVENT_KINDS, type MexConfig } from "mex-agent" — should resolve with full types.      

  Checklist

  - Tests pass (npm test) — 15 new tests green; 2 pre-existing failures unrelated to this PR.
  - No breaking changes — all additions are new exports, new files, or new optional parameters        
  defaulting to current behaviour. Internal callers (cli.ts, doctor.ts, tui.ts, watch.ts,
  sync/index.ts) are untouched and continue to compile.
  - Tested locally — typecheck, full test suite, and build all verified on Windows.

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,9 +1,0 @@
-{
-  "permissions": {
-    "allow": [
-      "Bash(git stash *)",
-      "Bash(npm test *)",
-      "Bash(npm run *)"
-    ]
-  }
-}
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -1,0 +1,9 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(git stash *)",
+      "Bash(npm test *)",
+      "Bash(npm run *)"
+    ]
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,6 @@ result
 
 # OS
 Thumbs.db
+
+# Per-user Claude Code config
+.claude/
--- a/package-lock.json
+++ b/package-lock.json
@@ -959,7 +959,6 @@
       "integrity": "sha512-F0R/h2+dsy5wJAUe3tAU6oqa2qbWY5TpNfL/RGmo1y38hiyO1w3x2jPtt76wmuaJI4DQnOBu21cNXQ2STIUUWg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "undici-types": "~6.21.0"
       }
@@ -970,7 +969,6 @@
       "integrity": "sha512-ilcTH/UniCkMdtexkoCN0bI7pMcJDvmQFPvuPvmEaYA/NSfFTAgdUSLAoVjaRJm7+6PvcM+q1zYOwS4wTYMF9w==",
       "devOptional": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "csstype": "^3.2.2"
       }
@@ -1508,7 +1506,6 @@
       "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "esbuild": "bin/esbuild"
       },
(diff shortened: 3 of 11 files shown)


theDakshJaitly:
Quick review — three small cleanups before merge:

1. **Drop `.claude/settings.local.json`** — that's per-user local config (Bash allowlist), looks accidental. Add it to `.gitignore` and remove from the diff.

2. **Clean the `package-lock.json` churn** — seven entries gain `"peer": true` (react, react-dom, esbuild, typescript, postcss, tsup, yaml) with no `package.json` change. Looks like an npm version difference on your machine rather than an intentional dependency change — please regenerate with the repo's npm version.

3. **Resolve `COMPATIBILITY.md` references** — `src/index.ts` and `test/public-api.test.ts` both reference "COMPATIBILITY.md at the repo root" but that file doesn't exist. Either add the doc or drop the references.

Small nits (non-blocking):
- `src/drift/index.ts:142` — `const scaffoldPatterns = patterns;` is a pointless rename now; just use `patterns` directly.
- The smoke test imports from `../src/index.js` (source), not the published bundle. Strong enough as a contract check, but a future `npm pack` + install test would catch packaging regressions this can't.

What's good: minimal public surface, `createConfig` guardrails on relative paths, `EVENT_KINDS` as single source of truth, signature changes are purely additive (no internal callers touched). All 153 tests pass locally.

CI note: `check (18)` was stuck on "Expected" because branch protection required it but the workflow matrix only runs `[20, 22]`. I've dropped 18 from protection — unrelated to this PR.

theDakshJaitly (approved):
LGTM — all three cleanup asks addressed and the pointless rename nit fixed. Public surface stays minimal and well-curated; CI green on Node 20 and 22; 153/153 tests pass locally. Approving and merging.

PR #58, opened 2026-06-02
fix(cli): read version from package metadata

### LINKED ISSUE #48: Fix hard-coded CLI version + add regression test
## The bug

`src/cli.ts:32` hard-codes the CLI version:

```ts
.version("0.3.5")
```

But `package.json` is at `0.5.0`, so `mex --version` reports the wrong number. This is the second version-string drift to ship, which is why this issue bundles a regression guard.

## Fix

Read the version from `package.json` so there is a single source of truth. Either:

- a tsup `define` / `__VERSION__` replaced at build time (preferred — no runtime JSON read), or
- a JSON import of `package.json` reading `.version`.

Either is fine; the constraint is **one source of truth = `package.json`**.

## Regression test

Add a test asserting `mex --version` output matches `package.json`'s `version` field, so this can't silently rot again.

## Acceptance criteria

- [ ] `mex --version` prints the version from `package.json` (currently `0.5.0`).
- [ ] A test fails if the reported version diverges from `package.json`.

Note: touches the tsup build config, so this is a notch above "good first issue."

ai-hustle-bro:
## Summary
- Read the CLI version from the root package.json instead of hard-coding it in src/cli.ts.
- Export CLI_VERSION and use it for commander --version wiring.
- Add a regression test that verifies the --version wiring uses package.json's version field.

Closes #48.

## Validation
- `npm run typecheck`
- `npm test -- test/cli.test.ts`
- `npm run build`
- `node dist/cli.js --version` -> `0.5.0`

Note: `npm test` currently has two Windows path-separator failures outside this change:
- `test/scanner.test.ts > scanEntryPoints > finds src/index.ts as main entry`
- `test/heartbeat.test.ts > heartbeat > flags files with stale last_updated frontmatter`

The new CLI version regression test passes in `test/cli.test.ts`.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1,7 +1,22 @@
 import chalk from "chalk";
 import { Command, InvalidArgumentError } from "commander";
+import { readFileSync } from "node:fs";
 import { findConfig } from "./config.js";
 import { reportConsole, reportQuiet, reportJSON, reportVerbose } from "./reporter.js";
+
+function readPackageVersion(): string {
+  const packageJson = JSON.parse(
+    readFileSync(new URL("../package.json", import.meta.url), "utf8"),
+  ) as { version?: unknown };
+
+  if (typeof packageJson.version !== "string" || packageJson.version.length === 0) {
+    throw new Error("package.json is missing a valid version field.");
+  }
+
+  return packageJson.version;
+}
+
+export const CLI_VERSION = readPackageVersion();
 
 export function parseIntArg(raw: string): number {
   const n = Number.parseInt(raw, 10);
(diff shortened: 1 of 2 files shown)


theDakshJaitly:
Thanks a lot for this, and sorry for the collision — you got caught by bad timing rather than anything wrong with your work. 🙏

Issue #48 was actually resolved and closed just minutes before this landed, in commit `fe9bddc` (shipped in **v0.5.1**). So `main` already reads the CLI version from `package.json` at runtime and has a matching regression test — your PR and that fix were in flight at the same time.

The merged fix lives in a dedicated `src/version.ts` module and resolves the path via `fileURLToPath` + `path.join` rather than `new URL("../package.json", import.meta.url)` — that's deliberate, so the bundler treats it as a plain runtime read instead of copying `package.json` as an asset. That's the main reason I'm not merging here, on top of the duplication.

One thing from your PR I genuinely liked: the explicit guard that throws on a missing/empty `version` field. The current `version.ts` doesn't have that, and it's a nice bit of defensiveness — happy to take it as a small follow-up if you'd like to open one.

Going to close this as already-resolved by #48 — not a reflection on the work, just unlucky timing. Really appreciate you digging in, and hope to see more PRs from you. 🚀

PR #60, opened 2026-06-03
Add TODO/FIXME drift checker and fix checker count docs

advancedresearcharray:
Closes #54
Closes #49

## Summary
- Adds `checkTodoFixme` — warns on unresolved `TODO` / `FIXME` markers in scaffold markdown (file + line).
- Registers the checker in drift detection with `TODO_FIXME` issue code and remediation hint.
- Updates README checker table, CONTRIBUTING checker count, and CHANGELOG [Unreleased].

## Test plan
- [x] `npm run typecheck && npm test && npm run build`

PR #61, opened 2026-06-04
feat: add broken-link drift checker

### LINKED ISSUE #52: Add a broken-link drift checker
## Context

mex ships 9 drift checkers in `src/drift/checkers/`, each a small self-contained function returning `DriftIssue[]`. This adds a 10th: a checker flagging Markdown links inside scaffold files whose target file doesn't exist.

Distinct from the existing `path` checker (`src/drift/checkers/path.ts`), which validates *path claims in prose* (e.g. "config lives in `src/config.ts`"). This checker targets Markdown link *targets* specifically — `[text](./relative/path.md)` where the target is missing.

## Scope

- Scan scaffold files for Markdown links `[...](...)` with local/relative targets.
- Flag links whose target file doesn't exist (resolve relative to the file containing the link).
- Skip external links (`http(s)://`, `mailto:`), anchors (`#section`), and code spans.

## Where

- New checker: `src/drift/checkers/broken-link.ts` — mirror `path.ts` (closest existing example: file-relative resolution).
- Register in `src/drift/index.ts` (import + invoke + push to `checkerIssueCounts`).
- Tests: `test/checkers.test.ts`.

## Acceptance criteria

- [ ] Flags a Markdown link to a non-existent local file.
- [ ] Ignores external links and anchors.
- [ ] Registered + counted in `src/drift/index.ts`.
- [ ] Tests cover broken / valid / external cases.

New to mex? Good first issue — 9 existing checkers to copy from. See CONTRIBUTING.md.

advancedresearcharray:
## Summary

Adds a **broken-link** drift checker for [#52](https://github.com/theDakshJaitly/mex/issues/52):

- Scans scaffold markdown for `[text](path)` links with local/relative targets
- Flags targets that do not exist (resolved relative to the containing file, then project/scaffold roots)
- Skips `http(s)://`, `mailto:`, `#anchors`, and links inside fenced or inline code

## Test plan

- [x] `npm run typecheck && npm test && npm run build`
- [x] Unit tests: broken link, valid link, external/anchor ignored, code-span ignored

Closes #52

### DIFF
--- a/src/drift/checkers/broken-link.ts
+++ b/src/drift/checkers/broken-link.ts
@@ -1,0 +1,97 @@
+import { existsSync, readFileSync } from "node:fs";
+import { dirname, resolve, relative } from "node:path";
+import type { DriftIssue } from "../../types.js";
+
+const LINK_RE = /\[([^\]]*)\]\(([^)]+)\)/g;
+
+/** Scan scaffold markdown for local links whose target file does not exist. */
+export function checkBrokenLinks(
+  scaffoldFiles: string[],
+  projectRoot: string,
+  scaffoldRoot: string
+): DriftIssue[] {
+  const issues: DriftIssue[] = [];
+
+  for (const filePath of scaffoldFiles) {
+    const source = relative(projectRoot, filePath);
+    let content: string;
+    try {
+      content = readFileSync(filePath, "utf-8");
+    } catch {
+      continue;
+    }
+
+    const fileDir = dirname(filePath);
+    const lines = content.split("\n");
+    let inFence = false;
+
+    for (let i = 0; i < lines.length; i++) {
+      const line = lines[i];
+      const trimmed = line.trim();
+      if (trimmed.startsWith("```")) {
+        inFence = !inFence;
+        continue;
+      }
+      if (inFence) continue;
+
+      const scanLine = line.replace(/`[^`]+`/g, "");
+      LINK_RE.lastIndex = 0;
+      let match: RegExpExecArray | null;
+      while ((match = LINK_RE.exec(scanLine)) !== null) {
+        const rawTarget = match[2].trim();
+        const target = normalizeLinkTarget(rawTarget);
+        if (!target || isExternalOrAnchor(target)) continue;
+
+        if (!linkTargetExists(target, fileDir, projectRoot, scaffoldRoot)) {
+          issues.push({
+            code: "BROKEN_LINK",
+            severity: "error",
+            file: source,
+            line: i + 1,
+            message: `Markdown link target does not exist: ${target}`,
+          });
+        }
+      }
+    }
+  }
+
+  return issues;
+}
+
+function normalizeLinkTarget(raw: string): string {
+  let target = raw.replace(/^<|>$/g, "").trim();
+  const titleSplit = target.match(/^([^\s]+)(?:\s+["'].+["'])?$/);
+  if (titleSplit) target = titleSplit[1];
+  return target;
+}
+
+function isExternalOrAnchor(target: string): boolean {
+  return (
+    /^https?:\/\//i.test(target) ||
+    /^mailto:/i.test(target) ||
+    target.startsWith("#")
+  );
+}
+
+function linkTargetExists(
+  target: string,
+  fileDir: string,
+  projectRoot: string,
+  scaffoldRoot: string
+): boolean {
+  const fromFile = resolve(fileDir, target);
+  if (existsSync(fromFile)) return true;
+
+  if (existsSync(resolve(projectRoot, target))) return true;
+
+  if (scaffoldRoot !== projectRoot && existsSync(resolve(scaffoldRoot, target))) {
+    return true;
+  }
+
+  if (target.startsWith(".mex/")) {
+    const withoutPrefix = target.slice(".mex/".length);
+    if (existsSync(resolve(projectRoot, withoutPrefix))) return true;
+  }
+
+  return false;
+}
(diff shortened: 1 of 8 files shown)


src/drift/checkers/broken-link.ts
```diff
@@ -0,0 +1,97 @@
...
+  }
+
+  return issues;
+}
+
+function normalizeLinkTarget(raw: string): string {
+  let target = raw.replace(/^<|>$/g, "").trim();
+  const titleSplit = target.match(/^([^\s]+)(?:\s+["'].+["'])?$/);
+  if (titleSplit) target = titleSplit[1];
+  return target;
```
theDakshJaitly, line 67: This is where the `#fragment` false positive originates. For `[install](./target.md#install)`, `target` stays `./target.md#install`, which fails `existsSync`, so an existing file is reported broken (at `error` severity). Reproduced locally.

Suggested fix — drop the fragment/query before returning:

```ts
function normalizeLinkTarget(raw: string): string {
  let target = raw.replace(/^<|>$/g, "").trim();
  const titleSplit = target.match(/^([^\s]+)(?:\s+["'].+["'])?$/);
  if (titleSplit) target = titleSplit[1];
  // Drop in-page fragment / query so links to a heading in another file resolve
  target = target.replace(/[#?].*$/, "");
  return target;
}
```

Bonus: a pure `#section` link then normalizes to `""` and is already skipped by the `!target` guard up in the loop, so same-page anchors stay ignored without needing the `startsWith("#")` branch. Worth a test for `[x](./target.md#heading)` (target exists) -> expect no issue.

src/drift/checkers/broken-link.ts
```diff
@@ -0,0 +1,97 @@
...
+      let match: RegExpExecArray | null;
+      while ((match = LINK_RE.exec(scanLine)) !== null) {
+        const rawTarget = match[2].trim();
+        const target = normalizeLinkTarget(rawTarget);
+        if (!target || isExternalOrAnchor(target)) continue;
+
+        if (!linkTargetExists(target, fileDir, projectRoot, scaffoldRoot)) {
+          issues.push({
+            code: "BROKEN_LINK",
+            severity: "error",
```
theDakshJaitly: Optional / non-blocking: `path.ts` (the closest analog) emits `warning` rather than `error` for `patterns/` files and placeholder-y paths, keeping `error` for genuinely-missing real paths. Since a broken link is the same class of problem, mirroring that would stop example links in pattern docs from costing 10 points each.

theDakshJaitly (changes requested):
Thanks for another checker, @advancedresearcharray — this one's noticeably more ambitious than #60, and it shows. The fenced-code and inline-code stripping is genuinely thoughtful (and you tested both), the file-relative + `.mex/`-prefix resolution mirrors `path.ts` correctly, and catching image links (`alt`) as a bonus is a nice touch. Tests pass and the structure is clean.

I'm marking this request-changes for two things before merge — both straightforward:

**1. Local links with a `#fragment` are falsely flagged as broken (see inline).** A link like `[install](./target.md#install)` reports `BROKEN_LINK` even when `target.md` exists, because the `#install` fragment is never stripped before the existence check. I reproduced it against the checker. Linking to a heading in another doc is a common pattern, and at `error` severity (-10 each) those false positives add up. The fix is small — strip a trailing `#...` (and ideally `?...`) in `normalizeLinkTarget`.

**2. Rebase on `main` and recount the checkers.** #60 just merged and touched the same README / CONTRIBUTING / CHANGELOG lines plus `index.ts` / `reporter.ts` / `types.ts` / `checkers.test.ts`, so this branch now conflicts on 6 files. After rebasing, `main` already lists **ten** checkers (it added `tool-config-sync` + `todo-fixme`), so with `broken-link` the count becomes **eleven** — the intro line should read "Eleven," and the table just needs the `broken-link` row added (the rest are already there). The CONTRIBUTING "Adding a drift checker" count moves to 11 too. (Tiny thing while you're in CHANGELOG: `main` uses `## [Unreleased]` with brackets to match the `## [0.3.5]` style — yours dropped the brackets.)

Non-blocking, optional follow-ups:
- **Severity:** `path.ts` downgrades to `warning` for `patterns/` files and placeholder paths, reserving `error` for real misses — worth considering the same here so a pattern doc demonstrating link syntax doesn't hard-error.
- Reference-style links (`[a][ref]`), paths with spaces, `~~~` fences, and double-backtick inline code aren't handled — all fine to leave; a one-line scope comment would be plenty.

Really solid work overall. Once the fragment fix and the rebase/recount are in, this is good to go. Thanks again for taking on the harder checkers.

advancedresearcharray:
Pushed fixes for the review in 73570d8: strip `#/?` from link targets before existsSync, add test for `./target.md#install`, and downgrade broken links under `patterns/` to warning (matching `path.ts`). Ready for another look when you have a moment.

advancedresearcharray:
Rebased onto `main` (post-#60) and resolved conflicts. Branch now includes both `todo-fixme` and `broken-link`; README/CONTRIBUTING/CHANGELOG updated to **11** checkers. Fragment stripping + `patterns/` warning severity unchanged in `d4d3d9b`. All 175 tests pass locally.

theDakshJaitly (approved):
Re-reviewed after the rebase — all good now.

- Fragment fix is in (`normalizeLinkTarget` strips `[#?].*$`). I re-checked independently: `./target.md#install` and query strings resolve, and a genuinely-missing file with a fragment still flags, so the strip isn't over-broad. Good regression test for it too.
- Rebased cleanly on `main`; checker count reconciled to eleven across the README intro, the table, and CONTRIBUTING, and the CHANGELOG `[Unreleased]` bracket style is preserved.
- You also picked up the optional severity nuance (`warning` in `patterns/`, mirroring `path.ts`) with a test — appreciated.

typecheck / 175 tests / build all green locally. Approving.

theDakshJaitly:
Merged — and this is your first contribution to land in mex. Thank you for taking on two of the trickier good-first-issues back to back, and for the clean turnaround here: the rebase, the fragment fix, and picking up the severity nuance unprompted all made this easy to ship. Genuinely solid work. Hope to see more from you whenever you're up for it.

PR #78, opened 2026-06-15
fix: make drift output and global config cross-platform (Windows)

Yashasvi2229:
## What

  Makes mex's drift output, global config, and test suite work correctly on
  Windows. Previously the new telemetry/feedback/drift code was only exercised
  on Linux CI, so several Windows-only bugs slipped through.

  Source fixes:
  - **Path separators.** `relative()`/`glob` return `\` on Windows, but drift
    issue `file` fields, heartbeat `staleFiles`, scanner entry paths, and
    event-log entries are forward-slash contracts (printed, JSON-serialized,
    consumed by mex-agent). The broken-link/path `patterns/` severity check
    (`source.includes("patterns/")`) silently misfired on Windows, producing the
    wrong severity. Added `src/paths.ts` `toPosix()` and applied it at every
    boundary.
  - **Home dir resolution.** `mexHomeDir()` relied on `homedir()`, which ignores
    `$HOME` on Windows (it reads `USERPROFILE`). Added a cross-platform
    `MEX_HOME` override so the global config / telemetry-id can be isolated and
    relocated.

  Test fixes:
  - telemetry/feedback suites now isolate via `MEX_HOME` instead of `$HOME`.
  - dev-repo guard tests `chdir` out before `rmSync` (Windows can't delete the cwd).
  - gated the POSIX `0o600` mode assertion to non-Windows (NTFS has no mode bits).

  ## Why

  On Windows the suite had 16 failing tests, and two were genuine product bugs
  (not just test issues): drift severity for `patterns/` files was computed
  incorrectly, and the global config/telemetry-id were written to the wrong
  directory. Linux CI could never catch these. This makes mex behave identically
  across platforms.

  ## Type of change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Docs
  - [ ] CI/Tooling

  ## How to test

  1. On Windows, run `npm install` then `npm test` → all 231 tests pass
     (previously 16 failed across `telemetry`, `feedback`, `heartbeat`,
     `scanner`, and `checkers`).
  2. On macOS/Linux, run `npm test` → still all green (no behavior change on POSIX).
  3. Spot-check the path fix: run `mex drift` on a project with a `patterns/*.md`
     file containing a broken link → it's reported as a **warning** (not an
     error) on every OS.

  ## Checklist

  - [x] Tests pass (`npm test`)

### DIFF
--- a/src/drift/checkers/broken-link.ts
+++ b/src/drift/checkers/broken-link.ts
@@ -1,5 +1,6 @@
 import { existsSync, readFileSync } from "node:fs";
 import { dirname, resolve, relative } from "node:path";
+import { toPosix } from "../../paths.js";
 import type { DriftIssue } from "../../types.js";
 
 const LINK_RE = /\[([^\]]*)\]\(([^)]+)\)/g;
@@ -13,7 +14,7 @@ export function checkBrokenLinks(
   const issues: DriftIssue[] = [];
 
   for (const filePath of scaffoldFiles) {
-    const source = relative(projectRoot, filePath);
+    const source = toPosix(relative(projectRoot, filePath));
     let content: string;
     try {
       content = readFileSync(filePath, "utf-8");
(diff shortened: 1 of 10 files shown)


theDakshJaitly (approved):
LGTM

PR #82, opened 2026-06-17
fix: skip HTML comments in checkBrokenLinks

Taegost:
## What

`checkBrokenLinks` now strips complete `<!-- ... -->` HTML comment spans before scanning for links. Previously, links inside HTML comments were flagged as broken — producing false `BROKEN_LINK` warnings in scaffold files like `patterns/INDEX.md` where template examples live in comment blocks.

Unclosed `<!--` (no matching `-->`) is left as plain text, consistent with `checkIndexSync`'s behavior.

**Line numbering fix:** The initial implementation stripped comments with an empty replacement, which collapsed line numbers for any real broken links appearing after a comment block. The replacement now preserves newline characters so the line array stays the same length as the original file — reported line numbers match the actual source location.

## Why

The `patterns/INDEX.md` template generated by `mex setup` includes example links inside HTML comments. After setup, running `mex check` on the scaffold produced 7 false `BROKEN_LINK` warnings from these comment examples — dropping the score to 79/100 even when the scaffold is otherwise correct.

Verified against [`Taegost/homelab-k8s` (`feat/mex-integration` branch)](https://github.com/Taegost/homelab-k8s/tree/feat/mex-integration).

## Changes

- `src/drift/checkers/broken-link.ts` — Strip `<!-- ... -->` from content before line processing, replacing with newline-preserving padding so line numbers remain accurate
- `test/checkers.test.ts` — 4 new tests: single-line comments, multi-line comments, inline comments with surrounding content, unclosed comments. Added `line` assertions to 2 existing tests. 2 regression tests verifying correct line numbers for broken links after single-line and multi-line HTML comments.
- `CHANGELOG.md` — Entry under `[Unreleased]`

The additions vs. the original:
- New "Line numbering fix" paragraph under "What" explaining the follow-up commit
- Updated the broken-link.ts bullet to mention newline-preserving padding
- Updated the checkers.test.ts bullet to mention the line assertions and regression tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,9 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+- **checkBrokenLinks false positives from HTML comments** — `checkBrokenLinks` now skips links inside HTML comments (`<!-- ... -->`), both single-line and multi-line. Previously, template examples in comments (e.g., in `patterns/INDEX.md`) were flagged as broken links.
+
 ## [0.6.1] - 2026-06-14
 
 ### Added
--- a/src/drift/checkers/broken-link.ts
+++ b/src/drift/checkers/broken-link.ts
@@ -24,7 +24,7 @@ export function checkBrokenLinks(
     const fileDir = dirname(filePath);
     // Strip complete HTML comments before line processing. Unclosed <!-- stays
     // as plain text (matches checkIndexSync behavior).
-    const lines = content.replace(/<!--[\s\S]*?-->/g, "").split("\n");
+    const lines = content.replace(/<!--[\s\S]*?-->/g, (m) => "\n".repeat(m.split("\n").length - 1)).split("\n");
     let inFence = false;
 
     for (let i = 0; i < lines.length; i++) {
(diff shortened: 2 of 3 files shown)


theDakshJaitly (changes requested):
Thanks for tracking down the HTML-comment false positives. The behavior change is right, but this implementation strips multi-line comments before splitting the file into lines, which changes the reported line numbers for any real broken links after a comment block.

For example, a link after:



would be reported a few lines earlier than its actual source location because the comment lines were removed before `i + 1` is assigned.

Could you preserve newline positions when removing complete HTML comments, or otherwise skip comment spans without changing the line numbering? A regression test with a real broken link after a multi-line comment would cover it.

Also, the PR is currently conflicting with `main`, so it needs an update before merge.

Taegost:
> Thanks for tracking down the HTML-comment false positives. The behavior change is right, but this implementation strips multi-line comments before splitting the file into lines, which changes the reported line numbers for any real broken links after a comment block.
> 
> For example, a link after:
> 
> would be reported a few lines earlier than its actual source location because the comment lines were removed before `i + 1` is assigned.
> 
> Could you preserve newline positions when removing complete HTML comments, or otherwise skip comment spans without changing the line numbering? A regression test with a real broken link after a multi-line comment would cover it.
> 
> Also, the PR is currently conflicting with `main`, so it needs an update before merge.

I updated the code to introduce `\n` to each line it removes, added tests, and updated the existing tests to validate the logic.

Issue #79, opened 2026-06-17
checkPaths produces false positives for non-path inline code in markdown


Taegost:
# Describe the bug

checkPaths extracts all backtick-wrapped strings from .mex/ markdown files and attempts to resolve each as a file path. Strings that don't resolve are reported as MISSING_PATH errors (−10 points each). This produces false positives for any backtick-wrapped string that isn't a file path — config values, commands, IPs, annotation keys, example filenames, file extensions, glob patterns, SMB shares, Windows paths, and general inline code formatting.

Backticks in markdown are general-purpose inline code formatting, not path declarations. The checker's assumption that all backtick-wrapped strings are path claims is incorrect for general markdown usage. On a correctly populated scaffold, mex check returns 0/100 because of these false positives, making the score meaningless.

# To Reproduce

1. Run npx mex-agent setup --mode agent-memory on any repository
2. Populate the scaffold context files with documentation that includes backtick-wrapped strings for non-path purposes, e.g.:
The Longhorn Helm value `csi.kubeletRootDir: /var/lib/kubelet` must not be changed.
The cluster subnet is `192.168.5.0/24`. The ArgoCD annotation `argocd.argoproj.io/sync-wave` controls sync ordering.
3. Run npx mex-agent check
4. See MISSING_PATH errors for every backtick-wrapped string that isn't a file path

# Expected behavior

checkPaths should only check paths that mex's own system depends on (e.g., frontmatter edges[].target values), not all backtick-wrapped strings in prose. Non-path inline code should not produce MISSING_PATH
errors.

# Command output

Drift score: 0/100 — 18 errors, 7 warnings, 0 info
12 files checked

ERROR

.mex/context/stack.md
  ✗ MISSING_PATH:80 csi.kubeletRootDir: /var/lib/kubelet

.mex/context/setup.md
  ✗ MISSING_PATH:180 csi.kubeletRootDir: /var/lib/kubelet
  ✗ MISSING_PATH:180 sudo ls /var/lib/kubelet/plugins_registry/
  ✗ MISSING_PATH:182 argocd.argoproj.io/sync-wave

.mex/context/decisions.md
  ✗ MISSING_PATH:31 192.168.5.0/24
  ✗ MISSING_PATH:38 forwardedHeaders.trustedIPs: 10.0.0.0/8
  ✗ MISSING_PATH:50 csi.kubeletRootDir: /var/lib/kubelet

.mex/context/conventions.md
[...]

PR #80, opened 2026-06-17
fix: scope checkPaths to ROUTER.md only

Taegost:
## What

Filters the claims array passed to `checkPaths` so it only includes claims extracted from `ROUTER.md`. This eliminates false `MISSING_PATH` errors from context docs, pattern files, and tool config files where backtick-wrapped strings are general-purpose inline code, not path declarations.

## Why

Fixes #79 — `checkPaths` was extracting every backtick-wrapped string from all scaffold markdown files and attempting to resolve each as a file path. Strings like `csi.kubeletRootDir: /var/lib/kubelet`, `argocd.argoproj.io/sync-wave`, and `192.168.5.0/24` were flagged as `MISSING_PATH` errors (-10 points each), producing a drift score of 0/100 on real-world scaffolds.

## Changes

- `src/drift/index.ts` — Added `basename` import; filtered `allClaims` to only `ROUTER.md` before passing to `checkPaths`
- `test/public-api.test.ts` — Added 3 integration tests verifying the scoping behavior
- `CHANGELOG.md` — Added entry under `[Unreleased]`

## Verification

Tested against https://github.com/Taegost/homelab-k8s branch `feat/mex-integration`  (the repo where the issue was reported):

| Metric | Before | After |
|--------|--------|-------|
| Score | 0/100 | **79/100** |
| Errors | 18 | **0** |
| Warnings | 7 | 7 |

The remaining 7 warnings are `BROKEN_LINK` issues in `patterns/INDEX.md` from a different checker — unchanged by this fix.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #83, opened 2026-06-18
fix: avoid path claims for non-path inline code

miloquinn:
## Summary
- Avoid extracting path claims from inline code values that are not filesystem paths
- Covers IP/CIDR values, file extensions, shell commands with path-like arguments, and dotted annotation/config keys
- Adds regression coverage based on the false-positive categories from #79

## Why
`checkPaths` currently treats many backtick-wrapped values as path claims when they contain `/` or a known extension. In project docs this creates noisy `MISSING_PATH` errors for Kubernetes/GitOps notes like `192.168.5.0/24`, `argocd.argoproj.io/sync-wave`, `sudo ls /var/lib/kubelet/plugins_registry/`, or `.yaml`.

Those are inline code snippets, not files the scaffold depends on.

## Testing
- `npm test -- test/claims.test.ts`
- `npm test -- test/checkers.test.ts`
- `npm run typecheck`
- `git diff --check`

### DIFF
--- a/src/drift/claims.ts
+++ b/src/drift/claims.ts
@@ -13,6 +13,18 @@ const TEMPLATE_PLACEHOLDER = /[<>\[\]{}]/;
 /** HTTP methods that indicate an API route, not a file path */
 const HTTP_METHOD_PREFIX = /^(GET|POST|PUT|PATCH|DELETE|HEAD|OPTIONS)\s+\//;
 
+/** IP addresses and CIDR ranges are network values, not filesystem paths. */
+const IP_OR_CIDR = /^(?:\d{1,3}\.){3}\d{1,3}(?:\/\d{1,2})?$/;
+
+/** Inline file extension references like `.yaml` describe a type, not a file. */
+const EXTENSION_ONLY = /^\.[A-Za-z0-9]+$/;
+
+/** Common shell commands that can contain path-like arguments. */
+const SHELL_COMMAND_PREFIX = /^(?:sudo\s+)?(?:ls|cd|cat|grep|find|kubectl|helm|docker|git)\s+/;
+
+/** Dotted config keys or annotations can contain slashes but are not paths. */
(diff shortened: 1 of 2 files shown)


theDakshJaitly (changes requested):
Thanks for the focused hardening here. The overall direction makes sense, especially after #80, since `ROUTER.md` can still contain inline values that look path-like but are not filesystem paths.

I found one false-negative risk before approving: `DOTTED_KEY_WITH_SLASH` also matches valid root hidden-directory paths like `.github/CODEOWNERS` and `.github/workflows`. Those are realistic scaffold references, and this change would stop extracting them as path claims, so `checkPaths` would no longer validate them.

Could you tighten that matcher so Kubernetes-style dotted annotation/config keys are skipped without excluding `.github/...` paths? A regression test for `.github/CODEOWNERS` or `.github/workflows` would cover it.

theyashasvipandey:
Thanks for this, and sorry it sat so long.

Your commit is now in `main` as `e430a8c`, with your authorship intact — it went in as the first commit of #170, which builds on it to close #107.

The one review point from @theDakshJaitly is addressed there rather than here. `DOTTED_KEY_WITH_SLASH` allowed the dotted segment to be empty, so alongside `argocd.argoproj.io/sync-wave` it also matched `.github/CODEOWNERS`, `.github/workflows` and `.mex/ROUTER.md` — which would have dropped mex's own scaffold paths out of `checkPaths` entirely. Requiring a real character before the dot keeps the Kubernetes-style keys skipped while leaving hidden directories checked, and the regression test he asked for is included.

The rest of #170 is the other half of #107: with the false positives gone from extraction, path checking could be widened from `ROUTER.md` to all eleven scaffold files. Your four patterns are load-bearing for that — a populated scaffold that scored 54/100 now scores 100/100 while checking eleven times more surface.

Closing this since the commit is merged. Thanks again.

PR #86, opened 2026-06-22
fix: detect and launch AI CLIs correctly on Windows (#85)

### LINKED ISSUE #85: mex sync interactive mode never detects installed CLI on Windows (uses which, spawns without shell)
## Environment
- OS: Windows 11
- mex-agent: 0.6.1
- Node: v24.16.0
- `claude` CLI installed and on PATH (`where claude` → `...\npm\claude.cmd`)

## Problem
On Windows, `mex sync` always prints:

```
No supported AI CLI detected. Falling back to prompts mode.
Supported CLIs: claude, opencode, codex
```

even though Claude Code is installed. Two Windows-specific bugs in `dist/cli.js`:

### 1. Detection uses `which`, which doesn't exist in cmd.exe
`hasCliTool()` and `hasClaudeCli()` run `execSync("which <cmd>")`. `execSync` uses cmd.exe on Windows, where the command is `where`, not `which`. It always throws → tool reported as not installed.

### 2. Launch uses `spawn`/`spawnSync` without `shell: true`
Even with detection fixed, `runToolInteractive()` (`spawnSync(meta.cli, args)`) and `launchClaude()` (`spawn("claude", ...)` with `shell: false`) fail with ENOENT on Windows, because the real binary is `claude.cmd` and Node won't resolve `.cmd` without a shell.

## Suggested fix
Make both platform-aware:

```js
// detection
execSync(`${process.platform === "win32" ? "where" : "which"} ${cmd}`, { stdio: "ignore" });

// launch
spawnSync(meta.cli, args, { ..., shell: process.platform === "win32" });
spawn("claude", [prompt], { ..., shell: process.platform === "win32" });
```

Affects `hasCliTool`, `hasClaudeCli`, `runToolInteractive`, and `launchClaude`.

Yashasvi2229:
## What

  Fix AI-CLI detection and launch on Windows. `mex sync` and
  `mex setup` never
  detected an installed AI CLI on Windows, so they silently
  fell back to
  copy-paste prompt mode even when Claude Code / Codex were
  installed.

  - **Detection** used `which`, which doesn't exist on Windows
  (the built-in is
    `where`). `execSync("which <cmd>")` always threw → every
  tool reported as not
    installed. Added `isCliAvailable()` (`src/cli-tools.ts`):
  probes with `where`
    on Windows, `which` elsewhere.
  - **Launch** used `spawn`/`spawnSync`, which throw `ENOENT`
  on the Windows
    `claude.cmd` wrapper. Switched the launch sites to
  `cross-spawn`, which
    resolves `.cmd`/`.bat` wrappers and escapes args correctly
  while keeping the
    multi-line prompt as a single argv element. (The naive
  `shell: true` is worse —
    Node concatenates args unescaped, DEP0190, shattering the
  prompt.)
  - **Robustness:** `runToolInteractive` treated `status ===
  null` (a spawn
    failure) as success, swallowing launch errors. Now returns
  false on
    `result.error` and only treats `status === 0` as success.

  ## Why

  Closes #85

  On Windows, the `which`-based detection made interactive
  sync/setup unusable —
  users were forced into copy-paste mode regardless of which
  AI CLI they had
  installed.

  ## Type of change

  - [x] Bug fix
  - [ ] New feature
  - [ ] Refactor
  - [ ] Docs
  - [ ] CI/Tooling

  ## How to test

  1. On Windows with `claude` installed and on PATH, run `mex
  sync` in a scaffold
     with drift → it now detects "Claude Code" instead of
  printing
     "No supported AI CLI detected. Falling back to prompts
  mode."
  2. Choose interactive (1) → Claude Code launches with the
  fix prompt (no
     `spawn claude ENOENT`).
  3. On macOS/Linux, behavior is unchanged. `npm test` → all
  240 tests pass.

  ## Checklist

  ## Checklist

  - [x] Tests pass (`npm test`)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -981,7 +981,6 @@
       "integrity": "sha512-F0R/h2+dsy5wJAUe3tAU6oqa2qbWY5TpNfL/RGmo1y38hiyO1w3x2jPtt76wmuaJI4DQnOBu21cNXQ2STIUUWg==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "undici-types": "~6.21.0"
       }
@@ -992,7 +991,6 @@
       "integrity": "sha512-ilcTH/UniCkMdtexkoCN0bI7pMcJDvmQFPvuPvmEaYA/NSfFTAgdUSLAoVjaRJm7+6PvcM+q1zYOwS4wTYMF9w==",
       "devOptional": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "csstype": "^3.2.2"
       }
@@ -1530,7 +1528,6 @@
       "dev": true,
       "hasInstallScript": true,
       "license": "MIT",
-      "peer": true,
       "bin": {
         "esbuild": "bin/esbuild"
       },
@@ -2679,7 +2676,6 @@
       "integrity": "sha512-QP88BAKvMam/3NxH6vj2o21R6MjxZUAd6nlwAS/pnGvN9IVLocLHxGYIzFhg6fUQ+5th6P4dv4eW9jX3DSIj7A==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=12"
       },
@@ -2729,7 +2725,6 @@
         }
       ],
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "nanoid": "^3.3.11",
         "picocolors": "^1.1.1",
@@ -2799,7 +2794,6 @@
       "resolved": "https://registry.npmjs.org/react/-/react-19.2.6.tgz",
       "integrity": "sha512-sfWGGfavi0xr8Pg0sVsyHMAOziVYKgPLNrS7ig+ivMNb3wbCBw3KxtflsGBAwD3gYQlE/AEZsTLgToRrSCjb0Q==",
       "license": "MIT",
-      "peer": true,
       "engines": {
         "node": ">=0.10.0"
       }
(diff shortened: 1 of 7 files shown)


theDakshJaitly:
Thanks for this — the diagnosis and fix are spot on, and switching to cross-spawn over `shell: true` is the right call. Nice that the new error handling also fixes the timeout-reported-as-success bug in `runToolInteractive`, not just the Windows path.

A few things before merge:

**1. Add a test for the `runToolInteractive` return-value logic.** It's the riskiest change in this PR and is currently untested — `isCliAvailable` has coverage, but the sync return logic doesn't. Mocking `crossSpawn.sync` to cover:

- `{ status: 0 }` -> `true`
- `{ status: 1 }` -> `false`
- `{ error: new Error(), status: null }` -> `false` (the timeout / ENOENT case)
- `{ status: null, signal: "SIGINT" }` -> `false`

would lock in the intended semantics.

**3. Trim the lockfile churn.** `package-lock.json` picked up a lot of unrelated `"peer": true` markers on existing packages (react, esbuild, typescript, postcss, vite-node, yaml, ...) beyond the cross-spawn additions — looks like a different npm version re-resolved the tree. Could you regenerate the lockfile with the project's npm version so the diff is limited to the cross-spawn entries, and confirm `npm ci` still resolves cleanly?

**5. Wrap the `launchClaude` call.** In `src/setup/index.ts`, `await launchClaude(prompt)` isn't in a try/catch, so a launch failure or a non-zero claude exit rejects and surfaces as an unhandled rejection. Since this PR is specifically about hardening the launch path, catching it to print a friendly message would round things out.

The other points from review (the user-interrupt behavior change, the `@types/cross-spawn` major-version lag) are non-blocking and fine as-is.

theDakshJaitly (approved):
LGTM

PR #101, opened 2026-07-15
feat(graph): add Rust code graph extractor

### LINKED ISSUE #96: [Rust] Add code-graph extractor
> Target branch: `code-graph-preview`

### Goal

Add a Rust extractor against the frozen `LanguageExtractor` interface. The initial implementation should model common modules and type declarations while treating macro-expanded behavior conservatively.

### Scope

- Detect `.rs` files as `rust`.
- Vendor and register a compatible Tree-sitter Rust WASM grammar.
- Emit a file node and containment tree.
- Extract functions, structs, enums and members, traits, implementations, methods, modules, constants, and static variables.
- Capture signatures, visibility, documentation comments, return types, and generic parameters when explicit.
- Emit `use`/module imports, calls, containment, trait implementation, type, and instantiation references where syntax is explicit.
- Leave cross-file and module targets unresolved through `targetName`.

### Required changes

- Add `src/graph/extraction/languages/rust.ts`.
- Register the extractor in `src/graph/extraction/languages/index.ts`.
- Register `.rs` and the grammar filename in `src/graph/extraction/grammars.ts`.
- Add the grammar WASM under `src/graph/wasm/` and document its source, version, and license.
- Add a representative Rust fixture and focused extractor tests.

### Acceptance criteria

- [ ] The extractor implements the frozen `LanguageExtractor` interface without changing it.
- [ ] Extraction is pure, deterministic, and limited to one file.
- [ ] Tests cover modules, functions, structs, enums, traits, `impl` methods, qualified names, containment, imports, calls, and explicit trait implementations.
- [ ] Macro invocations and unsupported constructs degrade conservatively without invented edges.
- [ ] Cross-file references remain unresolved for the engine to bind.
- [ ] `npm run typecheck`, `npm test`, and `npm run build` pass.
- [ ] The packed build contains the Rust grammar WASM.

### Out of scope

- Macro expansion or procedural macro execution.
- Cargo metadata execution or dependency downloading.
- Type inference and borrow-checker semantics.
- Changes to identity, reconciliation, schema, or grounding semantics.

### Contributor references

- [Extending the code graph](https://github.com/mex-memory/mex/blob/code-graph-preview/docs/extractors.md)
- `src/graph/extraction/languages/typescript.ts`
- `src/graph/__tests__/extractor.test.ts`

### Contributor workflow

```bash
git switch code-graph-preview
git pull
git switch -c feat/rust-code-graph-extractor
```

Open the pull request with base branch `code-graph-preview`.

abhinav-phi:
## What

- Added a new Rust code graph extractor implementing the existing frozen `LanguageExtractor` interface.
- Registered Rust support in the language and grammar registries.
- Added `.rs` file detection and vendored the Tree-sitter Rust WASM grammar.
- Added a representative Rust fixture and focused extractor tests covering common Rust constructs.

## Why

Closes #96

This adds initial Rust support to the code graph preview by extracting:

- Modules
- Functions
- Structs
- Enums
- Traits
- Impl blocks
- Methods
- Constants
- Static variables
- Imports (`use`)
- Calls
- Containment relationships
- Explicit trait implementations

Cross-file references remain unresolved through `targetName`, and macro invocations are handled conservatively as required.

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/Tooling

## How to test

1. Run:

   ```bash
   npm run typecheck
   ```

2. Run:

   ```bash
   npm test
   ```

3. Run:

   ```bash
   npm run build
   ```

4. Verify that:

   - `.rs` files are recognized correctly.
   - Rust modules, functions, structs, enums, traits, impl blocks, methods, constants, and statics are extracted.
   - Import, call, containment, and explicit trait implementation references are emitted.
   - Cross-file references remain unresolved through `targetName`.
   - The Rust grammar WASM is included in the build output.
   - The new Rust extractor tests pass.

## Checklist

- [x] Tests pass (`npm test`)
- [x] No breaking changes

## Code-graph preview

- [x] This PR targets `code-graph-preview`, not `main`
- [x] The linked issue states `Target branch: code-graph-preview`
- [x] The change follows the frozen `LanguageExtractor` or `FrameworkResolver` interface
- [x] A focused fixture and assertions for the expected node/edge shape are included
- [x] Any new grammar WASM, extension mapping, extractor, or resolver is registered
- [x] No graph identity, reconciliation, schema, or drift-semantics changes are included


src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,388 @@
...
+    if (!name) return;
+    const id = this.createNode("interface", name, node, {
+      visibility: visibilityOf(node),
+    });
+    if (!id) return;
+    const body = getChildByField(node, "body");
+    if (!body) return;
+    this.scopeStack.push(id);
+    for (const member of body.namedChildren) {
+      if (member.type === "function_item") {
```
theDakshJaitly: Trait methods without default bodies parse as `function_signature_item`, not `function_item`, so declarations such as `fn greet(&self) -> String;` are silently dropped. The current fixture contains exactly this case, but its test only asserts the `Greeter` interface node. Please handle signature-only trait methods too and assert that `Greeter::greet` is emitted with the expected containment/signature data.

src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,388 @@
...
+      for (const child of tokenTree.namedChildren) {
+         this.visit(child);
+      }
+    }
+  }
+
+  private walkBody(body: TSNode, ownerId: string): void {
+    const type = body.type;
+
+    if (CALL_TYPES.has(type)) {
```
theDakshJaitly, line 331: This body walker never handles Rust `struct_expression` nodes, so even the fixture's explicit `User { ... }` construction emits no `instantiates -> User` reference. Issue #96 explicitly requires instantiation references where syntax is explicit. Please extract struct construction here (and cover it with a focused assertion).

src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,388 @@
...
+      if (scope && scope.kind !== "file") parts.push(scope.name);
+    }
+    parts.push(name);
+    return parts.join("::");
+  }
+
+  private extractFunction(node: TSNode, isMethod = false): void {
+    const name = nameOf(node, this.source);
+    if (!name) return;
+    const id = this.createNode(isMethod ? "method" : "function", name, node, {
```
theDakshJaitly, line 153: Explicit generic parameters are not copied into `GraphNode.typeParameters`: for example, `fn make<T>` currently produces `typeParameters: undefined`. The same omission affects structs/enums/traits. Capturing explicit generics is an acceptance criterion in #96, so please parse the grammar's type-parameter field for each supported declaration and add coverage.

src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,388 @@
...
+    const typeName = typeNode ? getNodeText(typeNode, this.source) : "";
+    if (!typeName) return;
+
+    // We don't create a node for `impl` blocks themselves. They extend the type.
+    // Instead we map containment to the type if we can, but since target is just
+    // a name, we'll nest methods under a pseudo-scope so their qualified name is
+    // correct (e.g. `Foo::method`). 
+    
+    // Check if the struct/enum is already in the file.
+    let ownerId: string | null | undefined = this.nodes.find(n => n.name === typeName && (n.kind === "class" || n.kind === "enum" || n.kind === "interface"))?.id;
```
theDakshJaitly: This lookup makes containment depend on declaration order. Valid Rust such as `impl User { fn greet(&self) {} } struct User;` creates a namespace owner for the impl, and the method never becomes contained by the later `User` class node. Please make same-file type ownership order-independent, for example by collecting declarations before visiting impls or deferring impl binding.

src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,388 @@
...
+      } else if (fn.type === "scoped_identifier") {
+        calleeName = getNodeText(fn, this.source);
+      } else {
+        calleeName = getNodeText(fn, this.source);
+      }
+    }
+    if (calleeName) this.addRef(ownerId, calleeName, "calls", node);
+    
+    const args = getChildByField(node, "arguments");
+    if (args) for (const child of args.namedChildren) this.walkBody(child, ownerId);
```
theDakshJaitly: Arguments are walked here and then walked a second time when `walkBody` continues through all of the call node's `namedChildren`. As a result, `consume(make())` emits two identical `calls -> make` references, which are persisted as duplicate unresolved refs/edges. Please ensure call arguments are traversed exactly once and add a count-based regression assertion.

src/graph/extraction/grammars.ts
```diff
@@ -27,6 +27,7 @@ const WASM_GRAMMAR_FILES: Partial<Record<Language, string>> = {
   tsx: "tree-sitter-tsx.wasm",
   javascript: "tree-sitter-javascript.wasm",
   jsx: "tree-sitter-javascript.wasm",
+  rust: "tree-sitter-rust.wasm",
```
theDakshJaitly, line 31: The new vendored WASM binary is registered, but its upstream source, exact version/commit, and license are not documented anywhere in this PR. That provenance is explicitly required by issue #96 and `docs/extractors.md`; please add it alongside the vendored grammar documentation before merge.

theDakshJaitly (changes requested):
Requesting changes because the implementation does not yet satisfy several explicit acceptance criteria from #96. The existing typecheck, full test suite, and build pass, but focused cases reproduce the inline issues below.

abhinav-phi:
Addressed all requested review feedback:

- Added support for signature-only trait methods.
- Added struct instantiation extraction.
- Captured explicit generic type parameters.
- Made impl binding order-independent.
- Removed duplicate call edges.
- Documented the vendored Rust grammar provenance.
- Added focused regression tests for all requested cases.

Re-ran validation:

- ✅ `npm run typecheck`
- ✅ `npm test` (328/328)
- ✅ `npm run build`

src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,437 @@
...
+  }
+
+  private createNode(
+    kind: NodeKind,
+    name: string,
+    node: TSNode,
+    extra?: Partial<ExtractedNode>,
+  ): string | null {
+    if (!name) return null;
+    const id = generateNodeId(this.filePath, kind, name);
```
theDakshJaitly: Emitting signature-only trait methods now exposes a node-ID collision: `Greeter::greet` and `User::greet` both call `generateNodeId(filePath, "method", "greet")`, so they receive the same ID. The store upserts nodes by this primary key, causing one method to overwrite the other and corrupting the two containment relationships. Please add a persistence-level or ID-uniqueness regression and ensure both symbols can be represented without duplicate IDs. Because #96 freezes core identity changes, this may need maintainer coordination rather than silently persisting a lossy graph.

src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,437 @@
...
+    parts.push(name);
+    return parts.join("::");
+  }
+
+  private extractFunction(node: TSNode, isMethod = false): void {
+    const name = nameOf(node, this.source);
+    if (!name) return;
+    const id = this.createNode(isMethod ? "method" : "function", name, node, {
+      signature: signatureOf(node, this.source),
+      returnType: returnTypeOf(node, this.source),
```
theDakshJaitly: Issue #96 also requires explicit type references, but the extractor currently stores only the `returnType` string and emits no type edges. For example, `fn get() -> User` has no `returns -> User` edge, and a field `id: u32` has no `type_of -> u32` edge. Please emit the appropriate unresolved `returns`/`type_of` references where the grammar exposes these types, with focused assertions.

src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,437 @@
...
+      }
+    }
+    if (calleeName) this.addRef(ownerId, calleeName, "calls", node);
+    // Arguments are traversed naturally by walkBody via child traversal loop, avoiding duplication.
+  }
+
+  private extractInstantiation(node: TSNode, ownerId: string): void {
+    const nameNode = getChildByField(node, "name") ?? node.namedChild(0);
+    if (nameNode) {
+      const structName = getNodeText(nameNode, this.source);
```
theDakshJaitly: Using the raw node text leaves generic struct constructions unresolvable: `Boxed::<u8> { ... }` emits `targetName: "Boxed::<u8>"`, while the declared graph node is named `Boxed`. Please normalize this through a Rust-aware base-type helper so the `instantiates` reference targets `Boxed`, and add a generic-construction regression case.

docs/extractors.md
```diff
@@ -107,6 +107,16 @@ The resolver's language extractor must already be merged before a framework issu
 
 Use [`src/graph/__tests__/resolver-express.test.ts`](../src/graph/__tests__/resolver-express.test.ts) as the test pattern. Test positive and negative detection, the framework-specific node/reference shape, successful binding, and ambiguous or missing-target behavior.
 
+## Vendored Grammars
+
+When adding a language, document the grammar source and version.
+
+### Rust
+- **Official source:** `tree-sitter-wasms` package
```
theDakshJaitly: This version is not the source of the committed binary. The repository installs `tree-sitter-wasms` 0.1.12 in both `package.json` and `package-lock.json`, and the committed Rust WASM SHA-256 (`4409921a70d0aa5bec7d1d7ce809a557a8ee1cf6ace901e3ac6a76e62cfea903`) exactly matches `node_modules/tree-sitter-wasms@0.1.12/out/tree-sitter-rust.wasm`. Please correct the provenance/version (and distinguish the package license from the upstream grammar license if documenting both).

test/cli.test.ts
```diff
@@ -217,13 +217,20 @@ describe("built CLI main-module guard", () => {
 
   beforeAll(() => {
     execSync("npm run build", { cwd: repoRoot, stdio: "pipe" });
```
theDakshJaitly, line 219: This CLI timeout/Windows symlink behavior is unrelated to the Rust extractor or any requested review fix, and issue #96 scopes the contribution to the extractor, registrations, grammar, and focused tests. Please revert this unrelated test change or move it to a separately justified PR so this language-support change remains reviewable and scoped.

theDakshJaitly (changes requested):
Thanks for addressing the first review. The original regression cases now pass, as do typecheck, all 328 repository tests, and the build. A second graph-level pass found the inline blockers below. Additionally, `git diff --check` currently reports six trailing-whitespace errors in the updated test/extractor files; please clean those before merge.

abhinav-phi:
Thanks for the second review! I've addressed all the requested blockers:

- **ID collision**: Fixed by hashing the qualified name (`Owner::method`) for method IDs, while keeping the node name simple. Added a uniqueness test.
- **Type references**: Added unresolved `returns` edges for function/method return types and `type_of` edges for struct/enum fields.
- **Instantiation**: Stripped generics and turbofish syntax in `extractInstantiation` so constructions like `Boxed::<u8>` correctly emit an `instantiates` edge targeting the base `Boxed` node.
- **Grammar provenance**: Updated `docs/extractors.md` to reflect `tree-sitter-wasms` version `0.1.12` and matched the SHA-256 hash.
- **Trailing whitespaces**: Cleaned up all whitespace errors in the test and extractor files.
- **Unrelated changes**: Completely reverted the `cli.test.ts` Windows symlink changes to keep the PR strictly scoped.

All local validations (`npm run typecheck`, `npm test`, `npm run build`) and CI checks are fully green. Ready for another look!

src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,481 @@
...
+  }
+
+  private createNode(
+    kind: NodeKind,
+    name: string,
+    node: TSNode,
+    extra?: Partial<ExtractedNode>,
+  ): string | null {
+    if (!name) return null;
+    // For methods, use the qualified name (e.g. User::greet) as the ID basis to
```
theDakshJaitly: Using the qualified name as the hash input departs from the frozen `generateNodeId(filePath, kind, name)` identity contract, and it still does not fully solve Rust's legal collisions. If `Worker` implements both `Read::run` and `Write::run`, both implementation methods qualify as `Worker::run` and still receive the same ID; persistence overwrites one. Please add this multiple-trait-impl regression and coordinate the representation with maintainers rather than changing the identity input within this language extractor, since #96 explicitly excludes identity changes.

src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,481 @@
...
+    const name = nameOf(node, this.source);
+    if (!name) return;
+    const returnType = returnTypeOf(node, this.source);
+    const id = this.createNode(isMethod ? "method" : "function", name, node, {
+      signature: signatureOf(node, this.source),
+      returnType,
+      visibility: visibilityOf(node),
+      typeParameters: typeParametersOf(node, this.source),
+    });
+    if (!id) return;
```
theDakshJaitly, line 159: This emits the display text as the resolver target without reducing it to a symbol name. For example, `fn make() -> Boxed<User>` produces `returns -> "Boxed<User>"`, and the analogous field path produces `type_of -> "&User"`; neither can bind to the `Boxed` or `User` node. Please normalize explicit return and field types through an AST-aware base-type helper and cover generic/reference types, not only bare identifiers.

src/graph/extraction/languages/rust.ts
```diff
@@ -0,0 +1,481 @@
...
+    // Arguments are traversed naturally by walkBody via child traversal loop, avoiding duplication.
+  }
+
+  private extractInstantiation(node: TSNode, ownerId: string): void {
+    const nameNode = getChildByField(node, "name") ?? node.namedChild(0);
+    if (nameNode) {
+      // Strip generic turbofish syntax (`::<...>`) and angle-bracket generics
+      // so `Boxed::<u8> { ... }` resolves to `Boxed`, not `Boxed::<u8>`.
+      const rawName = getNodeText(nameNode, this.source);
+      const structName = rawName.replace(/::<[^>]*>/g, "").replace(/<[^>]*>/g, "").trim();
```
theDakshJaitly: Regex stripping is not balanced and breaks nested generic arguments: `Boxed::<Vec<u8>> { ... }` currently emits `targetName: "Boxed>"`. Please derive the constructor's base identifier from the syntax tree instead of matching angle brackets with `[^>]*`, and add a nested-turbofish regression assertion.

docs/extractors.md
```diff
@@ -107,6 +107,17 @@ The resolver's language extractor must already be merged before a framework issu
...
+## Vendored Grammars
+
+When adding a language, document the grammar source and version.
+
+### Rust
+- **Official source:** `tree-sitter-wasms` package, version `0.1.12`
+- **Vendored path:** `src/graph/wasm/tree-sitter-rust.wasm`
+- **SHA-256:** matches `node_modules/tree-sitter-wasms@0.1.12/out/tree-sitter-rust.wasm` exactly
+- **Upstream grammar:** [tree-sitter/tree-sitter-rust](https://github.com/tree-sitter/tree-sitter-rust)
+- **License:** MIT
```
theDakshJaitly: Please distinguish the two licenses here. The upstream `tree-sitter-rust` grammar is MIT, while the `tree-sitter-wasms@0.1.12` package identified as the binary source declares `Unlicense`. A single `MIT` label makes the provenance ambiguous and does not address the prior request to document both when they differ.

theDakshJaitly (changes requested):
The latest repository checks are green (`typecheck`, 332 tests, build, and `git diff --check`), but the remaining legal-Rust cases in the inline comments still produce lossy or unresolvable graph output, so I am keeping this on changes requested.

theDakshJaitly:
Thanks @abhinav-phi for the thorough follow-up — there is clear progress here. The basic trait-method extraction, declaration-order handling, duplicate-call fix, generic parameter capture, simple struct instantiation, grammar version correction, whitespace cleanup, and unrelated CLI-test revert all check out. I also reran the full validation locally: typecheck, all 332 tests, the build, and `git diff --check` are green.

I've left a small set of focused inline follow-ups for the remaining Rust identity and nested/generic type-resolution cases. Keeping the review on changes requested until those are resolved, but I appreciate the care put into addressing the earlier rounds.

src/graph/extraction/languages/rust.ts
theDakshJaitly: Resolved in `336cfc4` by restoring the frozen `generateNodeId(filePath, kind, name)` contract. The engine-wide same-named-symbol collision is intentionally deferred to #104 rather than changing identity inside this extractor.

docs/extractors.md
theDakshJaitly: Resolved in `336cfc4`. The documentation now distinguishes the `tree-sitter-wasms@0.1.12` binary package license (Unlicense) from the upstream `tree-sitter-rust` grammar license (MIT).

src/graph/extraction/languages/rust.ts
theDakshJaitly: Resolved in `336cfc4`. Struct instantiation now derives the base identifier from `generic_type_with_turbofish`; the regex was removed, and `Boxed::<Vec<u8>>` is covered by a focused regression.

src/graph/extraction/languages/rust.ts
theDakshJaitly: Resolved in `336cfc4`. Return and field type references now use an AST-aware base-type helper: generic returns target the outer symbol and reference fields target the referenced symbol. Focused regressions cover both cases.

theDakshJaitly (approved):
Re-reviewed final head 336cfc4 after the code-graph-preview rebase. All requested Rust extraction and provenance fixes are addressed; the extractor follows the frozen identity contract, with the engine-wide collision limitation tracked in #104. Typecheck, 343 tests, build, diff checks, packaging, and both CI jobs are green.

PR #105, opened 2026-07-22
Compact, budgeted code-graph retrieval + eval harness

theDakshJaitly:
Reworks the agent-facing code-graph surface so retrieval stops blocking a `main` release: it was dumping full source + full caller/callee arrays on every node with no budget or accounting, only ~1.3x leaner than grepping three files. This makes it a compact, budgeted, explainable, source-on-demand surface — and adds a committed eval harness that proves it.

## Results (this repo, dogfooded: 152 files, 1,801 nodes, 2,808 edges)

| metric | before | after |
|---|---|---|
| scope vs grep-top-3 (realistic median) | 1.34x | **10.06x** |
| scope vs whole corpus | ~122x | **969x** |
| expected symbol recall | 1.0 | **1.0** (held) |
| `runDriftCheck` over-expansion outlier | 0.26x / 32 facts | **5.90x / 9 facts** |

End-to-end with a real agent (opus-4-8, 5 NL tasks): both `--detail minimal` and `--detail source` answer 5/5 correctly. `minimal` is self-sufficient (0 Read/Grep fallback) via `graph get` round-trips; `source` is answer-ready but grep-falls-back ~1x/task. **Default `--detail` = `minimal`.**

## What's in it (M0–M5)

- **M0** deterministic graph reads (stable `ORDER BY` in `db/store.ts`) so a rebuilt graph is byte-identical — precondition for the harness.
- **M1** committed eval harness under `evaluate/` (`npm run eval`) — efficiency + search-quality, reproduces the prior ad-hoc benchmark bit-for-bit, hard CI gates.
- **M2** compact source-off retrieval: one JSONL protocol (`meta`/`fact`/`edge`/`source`/`summary`) with a hard token budget; source opt-in via `--detail source`; new `graph get <id>` verb. (Also: the 2 KB minhash fingerprint is no longer inlined on every fact — opt-in via `--fingerprint`, which grounding uses.)
- **M3** scored, quota-limited selection (`selectScope`): exact-identifier boosting, capped one-hop expansion, per-category quotas, `score` + `selectionReasons` on every fact.
- **M4** agent usage guidance in the persistent instruction templates ("treat returned source as already read", "don't re-scope on truncation", expand via `graph get`).
- **M5** end-to-end agent harness — scripted (`agent-e2e.mjs`) and real-model (`agent-e2e-model.mjs` via `claude -p`) — to settle the default `--detail`.

## Verification

- 360 unit tests green (`npm test`).
- `npm run eval` passes all gates (efficiency + search quality).
- Determinism: 3+ rebuilds produce byte-identical `scope`/`query`/`impact` output.
- Full benchmark writeup: `claude-talks/graph/GRAPH_RETRIEVAL_BENCHMARK_RESULTS.md`.

## Notes / follow-ups

- Small N (6 symbol + 5 NL tasks, one model, one repo); cost numbers are prompt-cache noisy. Default-detail call is defensible, not bulletproof — a larger fixture set would firm it up. Harness already supports `--root <other-repo>` for that.
- Deferred (non-blocking): flow-spine source selection + sibling skeletonization (would let `source` shed its fallbacks if ever promoted).


src/graph/agent-protocol.ts
```diff
@@ -0,0 +1,126 @@
...
+ */
+export class BudgetedEmitter {
+  private used = 0;
+  constructor(
+    private readonly write: (line: string) => void,
+    private readonly maxOutputTokens: number,
+    private readonly summaryReserve = 80,
+  ) {}
+
+  force(record: unknown): void {
```
theDakshJaitly: **P1 — the hard output-token ceiling can be exceeded.** `force()` bypasses the budget, the fixed 80-token reserve may be smaller than the summary, and `estimatedOutputTokens` is captured before the summary is added. Repro: `graph get missing:a missing:b missing:c --max-output-tokens 20` reports 72 estimated tokens, a 20-token maximum, and `truncated: false`. Please account for every emitted record (including errors and the footer), or reject budgets below a safe framing minimum.

src/graph/cli-agent.ts
```diff
@@ -81,40 +122,207 @@ export function runGraphQuery(
...
+      const fact = factFor(session, candidate.id, opts.detail, opts.fingerprint);
+      if (!fact) continue;
+      if (!emitter.offer({ type: "fact", ...fact, score: candidate.score, selectionReasons: candidate.reasons })) { truncated = true; break; }
+      returnedIds.add(candidate.id);
+      const node = session.graph.getNode(candidate.id);
+      if (node) returnedNodes.push(node);
     }
+
+    let returnedEdges = 0;
+    if (opts.detail !== "minimal") returnedEdges = emitInSetEdges(emitter, session.graph, returnedIds);
```
theDakshJaitly: **P1 — budget-dropped edges are reported as complete.** `emitInSetEdges()` only returns the number successfully emitted, so failed `offer()` calls never set `truncated`. Repro: `graph scope shallow --detail standard --max-output-tokens 700` returns all five facts, zero edges, and `truncated: false`; at a larger budget the same result contains three in-set edges. Source has a related issue: facts claim `sourceIncluded: true` before grouped source emission succeeds. Please propagate dropped-record state and only claim source inclusion when it was actually emitted.

src/graph/cli-agent.ts
```diff
@@ -38,22 +50,49 @@ export function runImpact(target: string, rootDir = process.cwd(), deps: AgentCo
...
 
-    writeJson(write, { type: "target", targetType: fileNodes.length > 0 ? "file" : "symbol", value: target });
+    const emitter = new BudgetedEmitter(write, opts.maxOutputTokens);
+    emitter.force(metaRecord("impact", opts));
+    emitter.force({ type: "target", targetType: fileNodes.length > 0 ? "file" : "symbol", value: target });
+
+    const emitted: GraphNode[] = [];
+    for (const root of roots.sort(byId)) {
+      const fact = factFor(session, root.id, opts.detail, opts.fingerprint);
+      if (fact) emitter.offer({ type: "defines", ...fact });
```
theDakshJaitly: **P2 — `impact --max-nodes` does not cap file-root definitions, and the summary overcounts returned nodes.** This ignores the result of `offer()` and pushes every root into `emitted`. Repro: `mex impact src/graph/cli-agent.ts --max-nodes 1` emits many definitions and reports `returnedNodes: 26`. Please apply the cap to the total returned set and only count nodes whose fact was successfully emitted.

theDakshJaitly (commented):
GitHub does not allow an author to request changes on their own PR, so this is submitted as a comment-state review. I recommend treating the retrieval protocol correctness issues called out inline as changes requested. The build, typecheck, tests, eval gates, and CI otherwise pass.

src/graph/cli-agent.ts
```diff
@@ -81,40 +122,207 @@ export function runGraphQuery(
...
-        continue;
-      }
+      if (relation === "where-defined") { resultNodes.push(node); continue; }
       const related = relation === "who-calls" ? session.graph.getCallers(node.id) : session.graph.getCallees(node.id);
-      for (const result of related.sort(byId)) {
-        writeJson(write, { type: "result", relation, target: node.id, ...hydrate(session, result, rootDir) });
-      }
+      for (const entry of related.sort(byId)) resultNodes.push(entry);
     }
+    const unique = dedupeById(resultNodes);
```
theDakshJaitly: **P2 — structural query results lose their target association.** The previous output included `target: node.id` on each result; flattening and deduplicating here means same-named/overloaded definitions can no longer be distinguished, and a caller related to two definitions is collapsed into one unexplained result. Please preserve `(target, result)` pairs and deduplicate by that pair.

theDakshJaitly:
Addressed all six review comments in 016dbe2 (thanks — these were real holes in the accounting layer, which is exactly the part that's supposed to be trustworthy). Each repro is now a regression test (+5 tests, 365 total green; eval gates still pass, median scope-vs-grep 10.57x, recall 1.0).

**P1 — hard ceiling exceeded / dishonest counts.** Replaced the emitter with a plan-then-emit `BudgetLedger`: framing (`meta`/`summary`) is counted rather than bypassed, `estimatedOutputTokens` now includes the summary, and `truncated` is true whenever any record is dropped or framing alone exceeds the budget. `graph get missing:a missing:b missing:c --max-output-tokens 20` now returns `truncated: true` with an honest token count.

**P1 — dropped edges / false sourceIncluded.** Budget-dropped edges now flip `truncated`. `sourceIncluded` is computed from a source-planning pass, so a fact only claims source when its source record actually fit; grouped-per-file source now degrades to per-range records so partial source still lands instead of dropping a whole file all-or-nothing.

**Copilot — `graph get --detail` unknown option.** Added `--detail` (fixed to source) to the `get` command. It was already documented in the templates, `suggestedNextCommands`, and the eval prompts, so this was broken guidance — it actually surfaced in the e2e run as a retry.

**P2 — impact overcounts / ignores cap.** `--max-nodes` now caps the total returned set (defines + callers) and only counts successfully-emitted nodes. `impact <file> --max-nodes 1` went from `returnedNodes: 26` to `1`.

**P2 — query loses target.** Restored `target` on each result and dedupe by `(target, result)` so overloaded/shared symbols stay distinguishable.

theDakshJaitly:
Round 2 addressed in e81a05f (both real — good catches).

**[P1] The ceiling is now genuinely hard, not just honest.** `beginResponse` sizes the summary reserve from the actual summary shape (its suggested commands carry long node ids/names, so the fixed 140 couldn't guarantee it) and clamps the budget up to a framing floor of one meta + one summary. So mandatory framing can never exceed the reported ceiling. `graph get missing:a missing:b missing:c --max-output-tokens 20` now reports `maxOutputTokens: 92` (clamped), `estimatedOutputTokens: 69` (<= 92), `truncated: true` — output never exceeds the reported budget, and the clamp is surfaced as truncation.

**[P2] Source planning no longer mutates already-accounted records.** `compactFact` now defaults `sourceIncluded: false`; the emitter flips it to true only for facts whose source actually fit — and true is a strictly shorter record than false, so the accounted shape is always >= the emitted one (ceiling stays safe). `runImpact` now passes only the fact records (defines/caller) into source planning, so the `target` record no longer gets a spurious `sourceIncluded`. Result: the reported estimate is no longer an under-count (your 224-vs-230 case now reports >= actual; measured 569 vs 566 emitted on `impact runImpact --detail source`).

+2 regression tests (367 total green): hard-ceiling clamp, non-fact records never stamped, and `estimatedOutputTokens` bounded by `[actual emitted, maxOutputTokens]` across budgets. Eval gates still pass (median 10.74x, recall 1.0), output deterministic.

The earlier review threads should be safe to resolve now — all six original findings plus these two are fixed and covered by tests.

PR #125, opened 2026-08-10
Skip non-scaffold files in tool config sync

olddognewflex:
## Problem

`checkToolConfigSync` treats any co-existing `CLAUDE.md` / `AGENTS.md` / `.cursorrules` / `.windsurfrules` / `.github/copilot-instructions.md` as byte-identical copies installed from `.tool-configs/`, and raises `TOOL_CONFIG_DRIFT` whenever two differ.

That assumption fails in repos where those files have independent owners. Real case: a repo with a hand-written `CLAUDE.md` (project context) and an `AGENTS.md` generated by `tkt sync-pack` (managed skill pack, "do not edit by hand"). Neither file ever came from `.tool-configs/`, and the warning is unfixable — making them match would either clobber the managed block or duplicate machine-regenerated content that re-drifts on the next sync.

## Fix

Every `.tool-configs/` template carries a `ROUTER.md` scaffold pointer. A file now participates in the mirror comparison only when its content includes that marker:

- Non-copies (no marker) are skipped entirely — no comparison, no warning.
- Genuine drift between real scaffold copies is still flagged, including when non-copies sit alongside them.

Trade-off: a copy edited so heavily that the `ROUTER.md` pointer is removed drops out of the check — at that point it is no longer a scaffold anchor, so silence is the correct behaviour.

## Testing

- Existing drift fixtures updated to carry the marker.
- Two new tests: hand-written + generated coexistence produces no issue; comparison still runs among the remaining real copies when non-copies are present.
- `npx vitest run test/checkers.test.ts` — 54/54 pass.
- Verified against the originating repo: `TOOL_CONFIG_DRIFT` false positive gone, drift score 88 → 91, real checks unaffected.

### DIFF
--- a/.mex/.tool-configs/.cursorrules
+++ b/.mex/.tool-configs/.cursorrules
@@ -4,7 +4,7 @@ description: Always-loaded project anchor. Read this first. Contains project ide
 last_updated: [YYYY-MM-DD]
 ---
 
-<!-- mex-tool-config: managed copy from .tool-configs/ -- keep this line so `mex drift` can detect out-of-sync copies -->
+<!-- mex-tool-config: managed copy from .tool-configs/ -- keep this line so `mex check` can detect out-of-sync copies -->
 
 # [Project Name]
 
--- a/.mex/.tool-configs/.cursorrules
+++ b/.mex/.tool-configs/.cursorrules
@@ -4,6 +4,8 @@ description: Always-loaded project anchor. Read this first. Contains project ide
 last_updated: [YYYY-MM-DD]
 ---
 
+<!-- mex-tool-config: managed copy from .tool-configs/ -- keep this line so `mex drift` can detect out-of-sync copies -->
+
 # [Project Name]
 
 ## What This Is
--- a/.mex/.tool-configs/.windsurfrules
+++ b/.mex/.tool-configs/.windsurfrules
@@ -4,7 +4,7 @@ description: Always-loaded project anchor. Read this first. Contains project ide
 last_updated: [YYYY-MM-DD]
 ---
 
-<!-- mex-tool-config: managed copy from .tool-configs/ -- keep this line so `mex drift` can detect out-of-sync copies -->
+<!-- mex-tool-config: managed copy from .tool-configs/ -- keep this line so `mex check` can detect out-of-sync copies -->
 
 # [Project Name]
 
--- a/.mex/.tool-configs/.windsurfrules
+++ b/.mex/.tool-configs/.windsurfrules
@@ -4,6 +4,8 @@ description: Always-loaded project anchor. Read this first. Contains project ide
 last_updated: [YYYY-MM-DD]
 ---
 
+<!-- mex-tool-config: managed copy from .tool-configs/ -- keep this line so `mex drift` can detect out-of-sync copies -->
+
 # [Project Name]
 
 ## What This Is
--- a/.mex/.tool-configs/CLAUDE.md
+++ b/.mex/.tool-configs/CLAUDE.md
@@ -4,7 +4,7 @@ description: Always-loaded project anchor. Read this first. Contains project ide
 last_updated: [YYYY-MM-DD]
 ---
 
-<!-- mex-tool-config: managed copy from .tool-configs/ -- keep this line so `mex drift` can detect out-of-sync copies -->
+<!-- mex-tool-config: managed copy from .tool-configs/ -- keep this line so `mex check` can detect out-of-sync copies -->
 
 # [Project Name]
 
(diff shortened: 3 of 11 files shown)


theDakshJaitly (changes requested):
The scaffold-copy detection remains too broad. ROUTER.md is not unique to mex-generated tool configs: independently owned CLAUDE.md and AGENTS.md files can legitimately tell agents to read ROUTER.md, causing this code to compare them and still report TOOL_CONFIG_DRIFT. I reproduced this with two unrelated files containing ordinary See ROUTER.md guidance. Please add a dedicated scaffold sentinel, such as an HTML mex-tool-config comment, to the generated templates and match that instead, with a regression test where non-scaffold files mention ROUTER.md. The focused tests, full suite (371 tests), and typecheck otherwise pass.

theDakshJaitly (dismissed):
Thanks for this — the diagnosis is right, and the second commit is the part I want to call out. Matching on `ROUTER.md` was too loose for exactly the reason you gave: an independently owned `CLAUDE.md` pointing agents at `ROUTER.md` is the normal case, not the exception. Good catch on your own patch, and the regression test for it is the right test.

I checked out the branch and ran the full suite: 372/372 pass across 37 files. Both invariants in `test/tool-config-templates.test.ts` still hold — the sentinel went into all four templates identically, so the "embedded configs are byte-identical" assertion and the "passes tool-config-sync after installation" assertion both survive.

One blocking issue before this can merge.

## The check goes silent for every existing install

The sentinel only exists as of this PR, and nothing puts it into copies that are already on disk. `src/setup/index.ts` skips any destination that already exists (`already exists — skipped`), so it will never rewrite someone's `CLAUDE.md`. There is no other code path that touches the installed root-level copies.

So for anyone who installed mex before this ships, every tool config fails the marker test, the checker skips all of them, and `TOOL_CONFIG_DRIFT` becomes unreachable. Verified on your branch:

```
CLAUDE.md      = pre-sentinel template text
.cursorrules   = same text + one hand edit   (genuine drift)

checkToolConfigSync() -> []     # 1 issue before this PR
```

Real drift, no warning, and the drift score rises as a result — so the only signal the user gets is the repo looking healthier than it is. That is the failure mode worth avoiding here: the false positive you fixed was at least visible, whereas this one is silent.

Worth noting the obvious fallback does not work. "If no file carries a sentinel, compare them all like before" reinstates the exact false positive this PR exists to fix, since the repo you hit it in has zero sentinels. The marker has to stay the sole positive signal, which means this needs a migration rather than a fallback.

## A fix that needs no migration step

Every tool-config template has carried this frontmatter line since the initial commit — I checked six vintages from the first release through `main`, and all four template variants:

```
description: Always-loaded project anchor. Read this first. Contains project identity, non-negotiables, commands, and pointer to ROUTER.md for full context.
```

Existing installs already carry a reliable proof of scaffold origin. It just is not the one the checker looks for. So accept either:

```ts
const SCAFFOLD_MARKER = "<!-- mex-tool-config";
// Copies installed before the sentinel shipped carry no marker, and nothing
// re-copies them. The template frontmatter has been byte-stable since the
// first release, so it identifies those copies without a migration step.
const LEGACY_MARKER = "Always-loaded project anchor. Read this first.";

const isScaffoldCopy = (c: string) =>
  c.includes(SCAFFOLD_MARKER) || c.includes(LEGACY_MARKER);
```

New installs match the sentinel. Existing installs match the fingerprint, with nothing to migrate. A hand-written `CLAUDE.md` or a `tkt sync-pack` `AGENTS.md` matches neither, so your false positive stays fixed — that description line is mex's own template prose and will not appear in an independently owned file.

Keep the sentinel as the forward contract; it is explicit, self-documenting, and survives a future rewrite of the frontmatter. The fingerprint is purely a bridge for installs already in the wild and can be dropped at a major version once those have turned over.

The caveat you already documented still applies and is still the right call: a file edited so heavily that both markers are gone drops out of the check. People edit the body sections rather than the frontmatter description, so this should be rare.

A test pinning the legacy path would be worth adding alongside the two you have — pre-sentinel content, genuine drift, one issue expected.

## Smaller things

- `templates/.tool-configs/README.md` line 4 still describes the contract as "most embed the same content" with no mention that a marker is now what enables the sync check. That README is where someone hand-maintaining these copies would look, and the sentinel's own inline text says the right thing already — the README should match it.
- `SCAFFOLD_MARKER` is an unterminated comment opener, so any file quoting the sentinel in prose reads as a copy. Low risk, but matching the closing `-->` or anchoring to line start costs nothing.

Separately, and not yours to fix: `present[0]` is the comparison reference, so when the first file is the drifted one every other file gets flagged instead. That predates this PR and I will open it as its own issue.

theyashasvipandey (approved):
Verified the legacy-marker path against real pre-sentinel template bytes: drift that this branch silently dropped is caught again, and the false positives stay fixed. CI green on 22 and 24.

Issue #110, opened 2026-08-01
graph: 'no such module: fts5' on Node 23.10.0 (inside documented engines range)


menelly:
## Summary

`mex graph` fails with `no such module: fts5` on Node 23.10.0 (Windows 11), which is inside the documented support range. `package.json` declares `engines: ">=22.5"` and COMPATIBILITY.md says 22.5+, but Node's *bundled* SQLite in this version isn't compiled with FTS5.

## Repro

```
mex graph
# -> no such module: fts5      (exits ~2s, no other output)
```

Confirmed it's Node's SQLite and not mex:

```js
const { DatabaseSync } = require('node:sqlite');
const db = new DatabaseSync(':memory:');
db.exec("CREATE VIRTUAL TABLE t USING fts5(x)");
// -> no such module: fts5     (node v23.10.0)
```

## Environment

| | |
|---|---|
| mex | 0.7.0 (built from source, `main`) |
| node | v23.10.0 |
| OS | Windows 11 |
| install | `npm install --ignore-scripts` + `npm run build` |

## Suggested fixes

1. **Narrow the supported range** in `engines` / COMPATIBILITY.md to whichever Node actually ships FTS5. I only tested 23.10, so I can't tell you where the real floor is — but 22.5 isn't it.
2. **Preflight check with an actionable message.** Right now it's one line with no hint, so it reads as a mex bug. Something like:
   `Your Node (v23.10.0) was built without SQLite FTS5, which the code graph requires. Upgrade to Node <X> or newer.`
   A cheap probe at startup: `CREATE VIRTUAL TABLE ... USING fts5()` in an in-memory db, catch, exit with the message.

## Doc note

COMPATIBILITY.md points users who can't upgrade Node at v0.6.3 — but 0.7.0 *is* the graph release, so that fallback drops the feature you'd be falling back for. Might be worth saying so explicitly.

## Two smaller things

**`.mex/` isn't gitignored.** The failed run left behind `.mex/graph.db` (4KB, broken) plus `-shm`/`-wal` files, untracked in my repo. Anyone who does `git add -A` will commit a broken database and two WAL files. Adding `.mex/` (or at minimum `*.db-wal` / `*.db-shm`) to `.gitignore` during setup would avoid a confusing diff.

[...]

PR #168, opened 2026-09-02
fix(graph): fail with an actionable message when Node's SQLite lacks FTS5

si-kui-a:
Fixes the crash-with-no-hint half of #110. `node:sqlite`'s bundled SQLite isn't guaranteed to be built with FTS5 on every Node build/version - not even inside the range `package.json`'s `engines` documents as supported, as you found on 23.10.0. Right now the first FTS5 statement in `schema.sql` throws SQLite's raw `no such module: fts5` uncaught, which reads like a mex bug rather than a Node/SQLite build limitation.

`openGraphDatabase` now probes FTS5 (create-and-drop a throwaway virtual table) right after opening the connection and before applying the schema, and turns that specific failure into `Your Node (vX.Y.Z) has SQLite built without FTS5 support, which mex's code graph requires. Try a different Node build/version - see COMPATIBILITY.md ...` - basically your own suggested probe and message. A non-FTS5 exec failure during the probe is re-thrown unchanged rather than misattributed.

Scoped to just that: I didn't touch the `engines`/COMPATIBILITY.md range (I only have your one data point, same as you noted), and left the `.gitignore`/telemetry-help-text notes from the issue for a separate pass since they're unrelated to this crash path.

Added `assertFts5Available` as its own exported function so it's unit-testable without needing an actual FTS5-less Node build: a fake `SqliteDatabase` covers the success path, the exact error string from your report, and confirms an unrelated exec failure isn't misattributed to FTS5. Also added a real `openGraphDatabase` test asserting the FTS5 tables still get created on a normal machine, so the preflight probe can't silently break the happy path.

`npm run typecheck && npm run test && npm run build` all clean, plus manually ran the built CLI's `mex graph` against a small fixture to confirm the preflight doesn't interfere with a real build.

https://claude.ai/code/session_01HG5wiw3Z8QdjNWb5bFDLiA


src/graph/db/database.ts
```diff
@@ -57,6 +57,33 @@ function configureReadOnlyConnection(db: SqliteDatabase): void {
...
+ * as supported (issue #110). Without this check, the first FTS5 statement in
+ * `schema.sql` throws SQLite's raw `no such module: fts5`, which reads like a mex
+ * bug rather than a Node/SQLite build limitation. Create-and-drop a throwaway
+ * virtual table rather than querying `pragma_module_list`, since that pragma is
+ * unavailable on some `node:sqlite` builds too and FTS5 usage is what actually
+ * needs to work.
+ */
+export function assertFts5Available(db: SqliteDatabase): void {
+  try {
+    db.exec("CREATE VIRTUAL TABLE IF NOT EXISTS __mex_fts5_probe USING fts5(x)");
```
theDakshJaitly: **[P1] Keep the capability probe off the persistent graph database**

Creating and dropping this virtual table changes the graph database bytes on every successful open. Existing read-oriented workflows use `openGraphDatabase` to obtain a graph, so `does not mutate the code graph by reading knowledge out of it` now fails locally and in Node 24 CI with a before/after hash mismatch. Please probe a temporary schema or separate in-memory connection (or use a genuinely non-mutating capability check) so opening an existing graph does not rewrite it.

src/graph/db/database.ts
```diff
@@ -82,6 +109,7 @@ export function openGraphDatabase(
   }
   const db = openSqlite(dbPath);
   configureConnection(db);
+  assertFts5Available(db);
```
theDakshJaitly: **[P1] Close the database when the preflight fails**

This newly throwing call is before the cleanup `try`, so a missing-FTS5 error leaves the `DatabaseSync` handle open. `GraphEngine` cannot close it because assignment only completes after `openGraphDatabase` returns; during `mex graph`'s isolated rebuild, Windows can then reject candidate deletion with `EPERM`, replacing the actionable FTS5 error in the exact environment from #110. Move configuration/preflight into the guarded block or explicitly close on failure.

theDakshJaitly (changes requested):
Requesting changes for two blocking regressions in the FTS5 preflight. The focused tests and typecheck pass, but the existing graph non-mutation regression fails locally and in Node 24 CI because the probe rewrites the graph database. The missing-FTS5 path also leaves the opened SQLite handle outside the existing cleanup guard, which can mask the intended diagnostic during Windows candidate cleanup.

src/graph/db/database.ts
si-kui-a: Thanks for catching this — `assertFts5Available` now probes against a throwaway `:memory:` connection instead of the caller's db. FTS5 support is a property of the SQLite build, not of any particular file, so nothing is lost by checking elsewhere. Confirmed the non-mutation regression test fails on the old code (hash mismatch) and passes with this change.

src/graph/db/database.ts
si-kui-a: Fixed — moved the `assertFts5Available()` call inside the existing try/catch, so a missing-FTS5 error now goes through the same `db.close()` cleanup as any other preflight failure. Added a test that mocks the preflight failure and asserts the handle is closed afterward.

theDakshJaitly (approved):
Re-reviewed commit fa4c350. Both requested changes are addressed: the FTS5 probe is isolated from the persistent graph database, and failure closes the graph handle. The focused regression coverage, local full verification, and the complete CI matrix are green. Approving.

Issue #112, opened 2026-08-03
[Flask] Add route resolver


theDakshJaitly:
> Target branch: `main`

## Dependency

The Python code-graph extractor was merged in #103 and is available on `main`.
This issue is not blocked by another language extractor.

## Goal

Add a bounded Flask framework resolver that connects decorator-defined routes
to their handler functions.

The initial implementation should cover conventional Flask application and
Blueprint decorators whose methods and paths can be determined statically.

## Scope

- Detect Flask from declared project dependencies such as `pyproject.toml`
  or requirements files.
- Recognize:
  - `@app.route("/path")`
  - `@app.route("/path", methods=["GET", "POST"])`
  - Shortcut decorators such as `@app.get()` and `@app.post()`
  - Equivalent decorators on Blueprint instances
- Treat `@app.route()` without an explicit methods list as a `GET` route.
- Emit a separate stable `route` node for each explicitly declared HTTP method.
- Emit a `function_ref` from each route node to the decorated handler.
- Resolve the reference only when the matching same-file function or method is
  unambiguous.
- Preserve Flask path converters such as `/users/<int:user_id>`.
- Leave missing or ambiguous handlers unresolved rather than guessing.

## Required changes

- Add `src/graph/resolution/frameworks/flask.ts` implementing
  `FrameworkResolver`.
- Register and export it from
  `src/graph/resolution/frameworks/index.ts`.
- Add a minimal Flask fixture under
  `src/graph/__tests__/fixtures/`.
- Add focused detection, extraction, resolution, and ambiguity tests.

## Acceptance criteria

- [ ] Positive and negative Flask detection are tested.
- [ ] Application and Blueprint route decorators are covered.
- [ ] Default `GET`, explicit method lists, and shortcut decorators are tested.
- [ ] Parameterized paths using Flask converters are preserved.
- [ ] Both synchronous and asynchronous handlers are covered.
- [ ] Route nodes use stable IDs and Python as their language.
- [ ] Same-file handlers resolve only when unambiguous.
- [ ] Missing and ambiguous handlers remain unresolved.
- [ ] The resolver is registered in the framework registry.
- [ ] `npm run typecheck`, `npm test`, and `npm run build` pass.
- [ ] No identity, reconciliation, schema, or drift-semantics changes are included.

## Out of scope

- Composing Blueprint prefixes across `register_blueprint()` calls.
- `add_url_rule()` and dynamically registered routes.
- `MethodView`, Flask extensions, middleware, and application factories.
[...]

PR #175, opened 2026-09-08
feat(graph): surface unindexed source files instead of leaving them silent

### LINKED ISSUE #163: graph: files in unsupported languages are silently absent — no warning at build, no signal at query
### `graph`: files in unsupported languages are silently absent — no warning at build, no signal at query

Thanks for mex — the scaffold and grounding halves have been solid for us. This is a
report about one failure mode in the graph half that is entirely silent, and I think the
silence is the bug rather than the language coverage.

#### Environment

mex-agent 0.7.3, node v26.0.0, macOS 26.6.2 arm64.

#### What happens

`EXTENSION_MAP` covers ten extensions — `.ts .mts .cts .tsx .js .mjs .cjs .jsx .py .rs` —
and `SUPPORTED_SOURCE_GLOB` is built from its keys. Every discovery path uses that glob:
`discoverSourceFiles`, `liveUnindexedFiles`, `findChangedSourceFiles`. A file with any
other extension is therefore invisible everywhere: not indexed, not a text fallback, not a
live-unindexed candidate.

Nothing reports this at any point.

**Repro A — a repo with no supported files at all.**

```bash
mkdir repro-go && cd repro-go && git init -q && mkdir .mex && echo '# Router' > .mex/ROUTER.md
cat > main.go <<'GO'
package main
type Server struct{ Addr string }
func NewServer(addr string) *Server { return &Server{Addr: addr} }
func (s *Server) Start() error { return nil }
GO
mex graph
```

```
Code graph built: 0 nodes, 0 edges across 0 files in 16ms → .mex/graph.db
```

Exit code 0. Then, from an agent's point of view:

| command | output |
|---|---|
| `mex graph scope "server start"` | `"status":"no-match"`, `"evidenceStrength":"none"`, `"warnings":[]` |
| `mex graph query where-defined NewServer` | `{"code":"TARGET_NOT_FOUND"}` |
| `mex impact Server` | `{"code":"TARGET_NOT_FOUND"}` |
| `mex doctor` | `✓ Drift 100/100`, four green ticks |

`TARGET_NOT_FOUND` is the same answer a typo gets. An agent concludes the symbol does not
exist and moves on. The only distinguishing signal is `health.indexedFiles: 0`, which
appears in `graph scope` output and not in `query` or `impact`.

**Repro B — a mixed repo, which is worse.** This is the common case, not an exotic one:
any Svelte, Vue or Astro frontend. (Your own field report in #140 describes a
"Python backend + TS/Svelte frontend".)

```bash
mkdir repro-mixed && cd repro-mixed && git init -q && mkdir -p .mex src
echo '# Router' > .mex/ROUTER.md
cat > src/api.ts <<'TS'
export function fetchOrders(userId: string): Promise<string[]> {
  return Promise.resolve([userId]);
}
TS
cat > src/OrderList.svelte <<'SV'
<script lang="ts">
  import { fetchOrders } from './api';
  export let userId: string;
  export function refreshOrders() { return fetchOrders(userId); }
</script>
<ul>{userId}</ul>
SV
cat > src/Checkout.vue <<'VUE'
<script setup lang="ts">
function submitPayment(amount: number) { return amount * 100; }
</script>
<template><button @click="submitPayment(1)">Pay</button></template>
VUE
mex graph
mex graph query where-defined refreshOrders
mex graph query where-defined submitPayment
mex graph query where-defined fetchOrders
mex graph scope "orders refresh"
```

```
Code graph built: 2 nodes, 1 edges across 1 files in 160ms → .mex/graph.db
{"type":"error","code":"TARGET_NOT_FOUND","target":"refreshOrders"}   # exists in the .svelte
{"type":"error","code":"TARGET_NOT_FOUND","target":"submitPayment"}   # exists in the .vue
... "name":"fetchOrders","filePath":"src/api.ts" ...                  # the .ts resolves fine
```

and scope reports:

```
"indexedFiles":1   "textFallbackFiles":[]   "warnings":[]
```

Three source files, one indexed. `indexedFiles` is non-zero, so even the single signal
from Repro A is gone. Two thirds of the codebase is missing from the graph and nothing
anywhere says so.

#### Why this matters more than coverage

The shipped `CLAUDE.md` template tells the agent to reach for the graph first and to treat
returned source as already read. When the graph is silently partial, the agent gets
`TARGET_NOT_FOUND` for symbols that exist and has no way to distinguish "not in this
codebase" from "this file type is not indexable". It then answers from a graph it believes
is complete.

Language coverage is tracked already (#93 for Go, #155 for C#) and I am not asking for
more extractors here. I am asking for the gap to be visible.

#### What would help

Any one of these would be enough:

1. `mex graph` reports skipped files: `Code graph built: 2 nodes ... across 1 files (2 files skipped: unsupported extension)`, with the extension histogram behind `--json`.
2. `filesIndexed` gains a sibling `filesSkipped` in the JSON, so a setup script can gate on it.
3. `query` / `impact` include `indexedFiles` (or a `coverage` field) the way `scope` does, so `TARGET_NOT_FOUND` can be told apart from `TARGET_NOT_INDEXABLE`.
4. `mex doctor` warns when a repo contains source files that no extractor handles.

Happy to supply more repros or test a patch.

abhinav-phi:
Resolves #163.

## Problem

`EXTENSION_MAP` covers ten extensions, and every discovery path is built from it, so a file with any other extension is invisible everywhere. The failure mode is silence:

- a repo with no supported files builds a graph of `0 nodes` with exit code 0, and `mex doctor` shows four green ticks;
- a mixed repo (any Svelte, Vue or Astro frontend) indexes one of three source files, reports `indexedFiles: 1` with no warning, and an agent asking about a symbol that exists gets the same `TARGET_NOT_FOUND` a typo gets.

## Approach

Count the complement at the one seam every consumer already shares. `corpus-policy.ts` gains `unindexedExtensionHistogram`: a walk with the same glob options and the same ignore list as source discovery (so a repository's own `graph.ignore` shapes the answer), keeping only files whose extension is recognized source but no extractor claims it.

The recognized set is checked through `isSupportedSourceFile` against the live extension map, so an extension stops being reported the moment an extractor lands for it — the list cannot go stale.

Reporting is bounded and best-effort by contract: a file-scan cap with an explicit `truncated` flag, a capped histogram, and a walk that can never fail the command it decorates.

## What changes

- **`mex graph`**: prints `Not indexed: N source file(s) have extensions no extractor handles: .go (2), ...` after the build summary; `--json` gains `unindexedSources` (`total`, `byExtension`, `truncated`), absent when the histogram finds nothing, so existing consumers are unchanged.
- **`mex graph query` / `mex impact`**: `TARGET_NOT_FOUND` records gain coverage context (`filesIndexed`, `unindexedSources`) — but only when it changes the record's meaning (nothing indexed, or unindexed source files exist). Ordinary misses in a healthy repository keep their exact prior shape.
- **`mex doctor`**: a `Coverage` line plus a next-steps hint when recognized source files are unindexable.

## Verification

- New focused tests: `coverage-histogram.test.ts` (recognition, noise exclusion, `graph.ignore` handling, truncation, entry cap).
- `cli-agent.test.ts`: two new tests — `TARGET_NOT_FOUND` stays bare on a fully covered repository, and names unindexed sources on a mixed TS/Svelte/Go repo.
- Local smoke test reproducing the issue's Repro B: mixed TS/Svelte/Go repo now reports `filesIndexed: 1` + `unindexedSources: {total: 2, byExtension: {".go": 1, ".svelte": 1}}` on both the build output and the `TARGET_NOT_FOUND` record; a clean repo's output is byte-identical to before.
- `npm run typecheck`, `npm run build` pass locally; targeted suites green. (The pre-existing Windows-local failures in symlink/SHM tests reproduce identically on clean `main` and are unrelated.)

No identity, reconciliation, schema, or drift-semantics changes are included.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Added
 - A bounded Next.js App Router resolver turning `app/**/route.ts|js` modules (including `src/app` roots) into route nodes: one per exported HTTP handler (`GET` through `HEAD`), with the URL path derived from the route file's directory, dynamic segments such as `[id]` and catch-alls preserved verbatim, and route groups `(marketing)` excluded the way Next resolves them. Same-file handlers resolve only when unambiguous; Pages Router, layouts, and pages stay out of scope (#95).
+- Coverage reporting for source files no extractor indexes. `mex graph` now prints the recognized-but-unindexed file count grouped by extension after the build summary, with the full histogram behind `--json` as `unindexedSources`; `mex graph query` and `mex impact` add `filesIndexed` and `unindexedSources` coverage context to `TARGET_NOT_FOUND` records (only when it changes the record's meaning, so misses in fully covered repositories are unchanged); and `mex doctor` shows a Coverage line. A mixed repository used to build a complete-looking graph while every `.svelte`, `.vue` or `.go` file was silently absent, indistinguishable from an empty one (#163).
 
 ### Fixed
 
--- a/src/doctor.ts
+++ b/src/doctor.ts
@@ -8,6 +8,7 @@ import {
 } from "./drift/index.js";
 import { checkHeartbeat } from "./heartbeat.js";
 import { readEvents } from "./events.js";
+import { unindexedExtensionHistogram } from "./graph/corpus-policy.js";
 import {
   graphChangeDetail,
   graphPrimaryDiagnostic,
@@ -25,9 +26,11 @@ export async function runDoctor(config: MexConfig): Promise<void> {
   const heartbeat = checkHeartbeat(config);
   const events = readEvents(config);
   const graph = report.graphStatus;
+  const coverage = unindexedExtensionHistogram(config.projectRoot);
 
   printLine("Drift", report.score >= 80 && errors === 0, `${report.score}/100 (${errors} errors, ${warnings} warnings)`);
   printLine("Graph", graph.status === "fresh", graphStatusDetail(graph));
+  printLine("Coverage", coverage.total === 0, coverageDetail(coverage));
   printLine("Heartbeat", heartbeat.ok, heartbeat.ok ? "HEARTBEAT_OK" : `${heartbeat.staleFiles.length} stale files, ${heartbeat.oldDailyMemoryFiles.length} old memory files`);
   printLine("Events", true, `${events.length} logged event${events.length === 1 ? "" : "s"}`);
   const hasConfig = existsSync(resolve(config.scaffoldRoot, "config.json"));
@@ -35,10 +38,13 @@ export async function runDoctor(config: MexConfig): Promise<void> {
 
   const graphNeedsAttention = graph.status !== "fresh";
   const graphRecoveryCommand = graphRemediationCommand(graph);
-  if (errors || warnings || !heartbeat.ok || graphNeedsAttention) {
+  if (errors || warnings || !heartbeat.ok || graphNeedsAttention || coverage.total > 0) {
     console.log();
     console.log(chalk.bold("Next steps"));
     if (errors || warnings) console.log("  Run `mex check` for drift details, then `mex sync` for targeted repair prompts.");
+    if (coverage.total > 0) {
+      console.log("  Source files above exist in the repository but are absent from the graph; no extractor ships for their language yet.");
+    }
     if (graphNeedsAttention && graphRecoveryCommand) {
       console.log(`  Run \`${graphRecoveryCommand}\` to repair the graph explicitly.`);
     } else if (graphNeedsAttention) {
@@ -48,6 +54,12 @@ export async function runDoctor(config: MexConfig): Promise<void> {
   }
 
   if (errors) process.exitCode = 1;
+}
+
+function coverageDetail(coverage: ReturnType<typeof unindexedExtensionHistogram>): string {
+  if (coverage.total === 0) return "all recognized source files are indexable";
+  const suffix = coverage.truncated ? ", walk stopped early" : "";
+  return `${coverage.total} source file(s) not indexable by any extractor${suffix}`;
 }
 
 function graphStatusDetail(graph: GraphAwareDriftReport["graphStatus"]): string {
--- a/src/graph/__tests__/cli-agent.test.ts
+++ b/src/graph/__tests__/cli-agent.test.ts
@@ -1342,6 +1342,64 @@ describe("runGraphQuery", () => {
     for (const result of results) expect(result).not.toHaveProperty("source");
   });
 
+  it("keeps TARGET_NOT_FOUND bare when every recognized source file is indexed", () => {
+    const bareRoot = mkdtempSync(join(tmpdir(), "mex-cli-agent-bare-"));
+    try {
+      writeFileSync(join(bareRoot, "only.ts"), "export const only = 1;\n");
+      const graph = syntheticScopeGraph({
+        nodes: [],
+        sources: [{ path: "only.ts", content: "export const only = 1;\n" }],
+        searchNodes: () => [],
+      });
+      const bareDeps: AgentCommandDeps = {
+        open: () => ({ graph, db: deps.open!(bareRoot).db, close: () => {} }),
+        write: (line) => lines.push(line),
+      };
+      const records = capture(() => runGraphQuery("where-defined", "ghost", bareRoot, bareDeps, {}));
+      const error = records.find((r) => r.type === "error" && r.code === "TARGET_NOT_FOUND");
+      expect(error).toMatchObject({ type: "error", code: "TARGET_NOT_FOUND", target: "ghost" });
+      expect(error).not.toHaveProperty("filesIndexed");
+      expect(error).not.toHaveProperty("unindexedSources");
+    } finally {
+      rmSync(bareRoot, { recursive: true, force: true });
+    }
+  });
+
+  it("names unindexed source files on TARGET_NOT_FOUND so a miss is not confused with an unindexable file", async () => {
+    const mixedRoot = mkdtempSync(join(tmpdir(), "mex-cli-agent-coverage-"));
+    let mixedEngine: GraphEngine | null = null;
+    let db: ReturnType<typeof openSqlite> | null = null;
+    try {
+      writeFileSync(
+        join(mixedRoot, "api.ts"),
+        "export function fetchOrders(userId: string): string[] {\n  return [userId];\n}\n",
+      );
+      writeFileSync(join(mixedRoot, "OrderList.svelte"), "<script lang=\"ts\">\nlet userId = '';\n</script>\n");
+      writeFileSync(join(mixedRoot, "main.go"), "package main\n");
+      mixedEngine = createGraphEngine({ rootDir: mixedRoot });
+      await mixedEngine.build(mixedRoot);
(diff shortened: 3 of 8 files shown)


theyashasvipandey:
Thanks @abhinav-phi,  right shape for #163: reporting rather than new extractors, bounded, and wrapped so it can never fail the command carrying it. The rebase note is accurate.

I merged current `main` into `8b30c9a` locally (only `CHANGELOG.md` conflicts now) and tested that: typecheck clean, your 53 tests pass, and I ran the built CLI over five fixtures plus a 30k-file tree.

Both of #163's repros are fixed for `mex graph`, `--json` and `TARGET_NOT_FOUND` from `where-defined` / `who-calls` / `impact`. A fully supported repo stays byte-identical, `node_modules` / `dist` / `graph.ignore` are excluded, freshness and `doctor`'s exit code are untouched, and candidates are checked against the live extension map so an extension drops off the moment an extractor lands.

## Must fix

### 1. The histogram re-walks the whole repository on every call

It globs `**/*` synchronously in `runGraph`, in `runDoctor`, and inside `targetNotFoundCoverage` — so on every `TARGET_NOT_FOUND` from `graph query` and `impact`, the path an agent hits most. On a 30k-file tree with **no unindexed sources at all**:

```
miss:  5.3 s   ->  {"type":"error","code":"TARGET_NOT_FOUND","target":"nothingHere"}
hit:   3.2 s
```

It walks everything, finds nothing, discards the result, and emits exactly what it emitted before. The repos paying most are the ones the feature does nothing for.

A second problem rides along: `maxUnindexedFiles: 20_000` counts every **scanned** file, not candidates (the build's own `maxSourceFiles` counts source files only), so any repo over 20k non-ignored files always truncates. My 30k tree holds 600 `.go` files and reports:

```
Not indexed: 399 source file(s) have extensions no extractor handles: .go (399)
  (count may be higher: the repository walk stopped early)
```

399 of 600, and which 399 depends on glob order.

The fix I'd like is to compute it once during the build and have the read paths reuse it — and I want to be honest that it isn't a small change. `project_metadata` is already in the shipped schema (`schema.sql:288`) and the build already writes through it (`engine-impl.ts:535`), so no schema change is needed, and `storedManifestIdentity` (`cli-agent.ts:2229`) is already the read pattern. But `doctor` never opens the graph store, and a stored count goes stale as soon as files change without a rebuild, so it has to hang off the freshness machinery. If you'd rather not take that on in this PR, say so and I'll split it into its own issue — but then the walk needs a much tighter bound before this merges. Either way, please cap on candidate files rather than on everything scanned.

### 2. `graph scope` still reports everything is fine

#163's Repro B, and `scope` is where the shipped agent guidance sends agents first. Mixed repo, 1 `.ts` + 1 `.svelte` + 1 `.vue`:

```
{"type":"health","indexedFiles":1,"okFiles":1,"partialFiles":0,"failedFiles":0,"staleFiles":[]}
{"type":"summary","matchedNodes":1,…,"status":"ok","evidenceStrength":"moderate","warnings":[],…}
```

Two thirds of the code is invisible and the answer is `ok` / `moderate` / no warnings. The `health` record already carries `indexedFiles` (`cli-agent.ts:291`), so it's the natural home. Heads-up before you write it: `HealthRecord` is part of `AGENT_PROTOCOL_VERSION = 3` and `fixtures/protocol-v3/graph-scope.jsonl` freezes that output byte for byte — I'll confirm whether an additive field is fine inside v3 before you update the golden.

### 3. CHANGELOG

Only `CHANGELOG.md` conflicts now. `main` gained the Flask (#112) and NestJS (#98) entries plus the Hub tour under `## [0.8.2] - Unreleased` → `### Added`; keep all of them and add yours to the same list.

## Should fix

- **`graph rebuild` and `graph refresh` stay silent.** `printUnindexedSources` is only called from `runGraph`, so `rebuild` prints no coverage line and `rebuild --json` has no `unindexedSources` key — and those are the commands `doctor` tells people to run after a graph problem.
- **Extension noise.** `OTHER_KNOWN_SOURCE_EXTENSIONS` includes `.sql`, `.proto`, `.gradle`, `.ps1`, `.bash`, `.zsh`, `.vim`. This repo trips its own check: `mex doctor` prints `! Coverage 1 source file(s) not indexable by any extractor`, and that file is `src/graph/schema.sql`, our own DB schema. I'd drop the extensions no extractor will plausibly claim and keep the actual languages.
- **`doctor`'s next step points at nothing.** It says "Source files **above** exist in the repository but are absent from the graph", but `doctor` prints only a count — no extensions, nothing above it. Either put the top extensions on the `Coverage` line (same `entries` array `mex graph` formats) or point the hint at `mex graph --json`.

## Smaller

- `TARGET_NOT_FOUND`'s `unindexedSources` omits `truncated`, unlike `--json`, so a truncated count reads as exact.
- `mex impact src/Checkout.vue` could say *this* file is unindexable rather than leaving the agent to infer it from a repo-wide count.
- `vendor/` isn't ignored, so vendored Go/PHP/Ruby deps get counted — consistent with the build's ignore list, just worth knowing.

abhinav-phi:
Implemented the build-time cache in a6ae73d/a4c2fff using existing project_metadata. Reads reuse it with bounded directory-stamp validation, without a coverage re-walk. The 30k-file regression counts all 600 Go candidates; stale, malformed, or over-budget observations do not supply exact totals. Build/refresh/rebuild output, doctor guidance, extension filtering, and the miss truncated flag are updated. The changelog conflict is resolved with the upstream entries retained.

Scope now warns and degrades for observed unsupported coverage. An invalidated cache produces an explicit unknown-coverage warning while preserving the status semantics of returned evidence. HealthRecord, protocol v3, and goldens are unchanged pending your approval of any additive health field. The optional target-specific unsupported-file message is not included.

Final-head CI is green: https://github.com/mex-memory/mex/actions/runs/35164795895 . @theyashasvipandey could you review the cache/freshness approach and confirm whether you still want the additive HealthRecord field?

abhinav-phi:
hi the coverage cache is implemented in a4c2fff

coverage is computed once at graph build time and stored in the existing metadata

reads reuse the stored count with bounded directory stamps and never rewalk the repository

a 30k file tree now counts all 600 go candidates exactly

stale malformed or over budget observations never report an exact total

graph scope warns and degrades on observed unsupported coverage and shows an unknown coverage warning when the cache is stale

the changelog conflict is resolved with all upstream entries kept

the health record and protocol goldens stay untouched pending your call on additive fields

all ci checks are green

could you review the cache approach and let me know if you still want the additive health field

theyashasvipandey:
Thanks @abhinav-phi! To your questions:

- **Cache/freshness approach:** approved, and I kept it.
- **Additive `HealthRecord` field:** not wanted. `scope` warnings are enough, and protocol v3 stays untouched.
- **Message naming the specific file:** fine to skip.

pushed two commits.

- **`44caadc`** merges `main` and resolves the `doctor.ts` conflict with #181.
- **`365aa49`** has three fixes:
  - **Faster reads.** Stamps are checked with `lstat` instead of `realpath`, and skipped when there's nothing to report. On a 921-directory repo, `scope` went from ~0.7 s to ~4 ms.
  - **Stale counts are labeled, not dropped.** An unrelated change like a README save used to make a miss bare again. Unverified counts now say `observedAt: "last-build"`, and unreadable ones say `coverage: "unknown"`.
  - **Coverage no longer turns `scope` answers `degraded`.** On real multi-language repos, test fixtures in other languages made 10 of 12 strong answers `degraded`. The warning now shows only when evidence is weak or empty.

Merging.

PR #178, opened 2026-09-08
test(cli): smoke-test the built CLI against a fixture scaffold

### LINKED ISSUE #57: Add a CLI smoke test against a fixture scaffold
## Context

Two version bugs shipped because CI never exercises the CLI end-to-end. Add a smoke test that runs each top-level command against a fixture scaffold and asserts exit codes + basic output. This would have caught the hard-coded-version bug.

## Scope

- Build a fixture scaffold (commit one, or generate via `mex setup` in a temp dir).
- Run each top-level command (`check`, `doctor`, `timeline`, `log`, `--version`, …) against it.
- Assert exit code `0` and key output substrings for each.

## Where

- New test, e.g. `test/cli-smoke.test.ts`.
- CI already runs `npm test` (`.github/workflows/ci.yml`), so wiring is automatic once the test exists.

## Acceptance criteria

- [ ] Smoke test runs all top-level commands against a fixture scaffold.
- [ ] Asserts exit codes and key output for each.
- [ ] A bad `--version` (e.g. mismatched constant) would fail the suite.

abhinav-phi:
Resolves #57.

## What

A smoke suite that spawns the built `dist/cli.js` in a temp project and asserts exit codes and output for the commands a user or agent reaches for first:

- `--version` — strict equality against `package.json`'s version
- `log` → `timeline` → `timeline --json` — one event rendered end to end in both output modes
- `check` and `check --json` — healthy fixture reports `100/100`, JSON is parseable with the documented shape
- `heartbeat`, `doctor` — run clean on the fixture
- `--help` — lists the top-level commands

## Fixture

Each test gets a fresh temp project with a minimal `.mex/` scaffold: `ROUTER.md` plus an `AGENTS.md` tool anchor — the state `mex setup` leaves behind. The anchor matters: without it, `check` reports `SCAFFOLD_ORPHANED` (exit 1), which is correct product behavior and exactly the kind of drift the smoke test should be able to express, not fight.

## Why this would have caught the shipped bugs

The two version regressions shipped because CI exercised library code and the packed install, but never ran the real CLI against a scaffold inside the suite. I verified the detection property directly: injecting a hard-coded `VERSION = "0.7.0"` into the bundle (reproducing the original hard-coded-literal bug) makes the `--version` test fail with `expected '0.7.0' to be '0.8.0'`. CI already runs `npm run build` before `npm test`, so the suite has the bundle it needs; the test also fails with an explicit message if `dist/cli.js` is missing.

No production code is touched — test + fixture only.

### DIFF
--- a/test/cli-smoke.test.ts
+++ b/test/cli-smoke.test.ts
@@ -1,0 +1,149 @@
+import { describe, it, expect, beforeAll, afterAll } from "vitest";
+import { spawnSync } from "node:child_process";
+import { existsSync, mkdirSync, mkdtempSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+
+// End-to-end smoke test for the shipped CLI against a real scaffold fixture.
+// Two version bugs shipped because CI only exercised library code and the
+// packed-install smoke: the CLI itself was never run against a scaffold in the
+// suite. This spawns the built `dist/cli.js` for every top-level command a
+// user or agent reaches for first, asserting exit codes and the output that
+// makes each command worth running.
+
+const repoRoot = join(dirname(fileURLToPath(import.meta.url)), "..");
+const cliPath = join(repoRoot, "dist", "cli.js");
+const pkg = JSON.parse(readFileSync(join(repoRoot, "package.json"), "utf8")) as { version: string };
+
+const roots: string[] = [];
+
+beforeAll(() => {
+  // The suite's other CLI tests build dist/ themselves; the smoke test depends
+  // on it, so fail with the reason instead of a confusing MODULE_NOT_FOUND.
+  if (!existsSync(cliPath)) {
+    throw new Error("dist/cli.js is missing — run `npm run build` before the smoke test.");
+  }
+});
+
+function project(): string {
+  const root = mkdtempSync(join(tmpdir(), "mex-cli-smoke-"));
+  roots.push(root);
+  return root;
+}
+
+function scaffold(root: string, files: Record<string, string> = {}): void {
+  const mexDir = join(root, ".mex");
+  mkdirSync(mexDir, { recursive: true });
+  writeFileSync(join(mexDir, "ROUTER.md"), "# Router\n\nEntry point for the scaffold.\n");
+  // A populated scaffold without a tool anchor reports SCAFFOLD_ORPHANED —
+  // the fixture mirrors what `mex setup` produces: an agent entry point.
+  writeFileSync(join(root, "AGENTS.md"), "# Agents\n\nUse the scaffold under .mex/ as the project memory.\n");
+  for (const [path, content] of Object.entries(files)) {
+    mkdirSync(dirname(path) === path ? mexDir : join(root, dirname(path)), { recursive: true });
+    writeFileSync(join(root, path), content);
+  }
+}
+
+function mex(
+  root: string,
+  args: string[],
+): { status: number; stdout: string; stderr: string } {
+  const result = spawnSync(process.execPath, [cliPath, ...args], {
+    cwd: root,
+    encoding: "utf8",
+    env: {
+      ...process.env,
+      HOME: root,
+      MEX_TELEMETRY: "0",
+      DO_NOT_TRACK: "1",
+      NO_COLOR: "1",
+    },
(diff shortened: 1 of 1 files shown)


test/cli-smoke.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    cwd: root,
+    encoding: "utf8",
+    env: {
+      ...process.env,
+      HOME: root,
+      MEX_TELEMETRY: "0",
+      DO_NOT_TRACK: "1",
+      NO_COLOR: "1",
+    },
+  });
```
theDakshJaitly, line 67: [P2] Bound child-process execution

`spawnSync` has no `timeout`, so a CLI regression that never exits can block this worker beyond Vitest's test timeout. In a controlled probe, a three-second child delay with a one-second Vitest deadline reported failure only after the child returned (3.39 seconds). Please add a finite subprocess timeout and include `result.error`/signal and stderr in the failure diagnostic, so a hung command becomes an actionable test failure.

test/cli-smoke.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    const timeline = mex(root, ["timeline"]);
+    expect(timeline.status).toBe(0);
+    expect(timeline.stdout).toContain("smoke: chose the bounded resolver");
+
+    const asJson = JSON.parse(mex(root, ["timeline", "--json"]).stdout) as {
+      events: Array<{ kind: string; message: string }>;
+    };
+    expect(asJson.events).toEqual([
+      expect.objectContaining({ kind: "decision", message: "smoke: chose the bounded resolver" }),
+    ]);
```
theDakshJaitly: [P2] Assert the JSON timeline exit code

This parses stdout and discards the process status. Forcing only `timeline --json` to return exit code 1 while retaining its valid JSON output still passed all seven smoke tests. That misses the exit-code regression this suite is meant to catch under #57. Please retain the result and assert `status === 0` before parsing stdout.

test/cli-smoke.test.ts
```diff
@@ -0,0 +1,149 @@
...
+    expect(report.score).toBe(100);
+    expect(Array.isArray(report.issues)).toBe(true);
+  });
+
+  it("heartbeat runs clean on the fixture", () => {
+    const root = project();
+    scaffold(root);
+
+    const result = mex(root, ["heartbeat"]);
+    expect(result.status).toBe(0);
```
theDakshJaitly, line 143: [P2] Assert heartbeat's success output

This only checks exit status. Replacing `heartbeat` with an immediate successful exit and no output still passed all seven smoke tests, so a missing or no-op command handler goes undetected. Please also assert that stdout contains `HEARTBEAT_OK`, matching #57's requirement to check both exit codes and key output.

theDakshJaitly (changes requested):
Requesting three small test-harness fixes before merging: bound the child process, assert the JSON timeline exit status, and assert heartbeat's success output.

I tested the exact PR test file on main at 9648c9d. All seven smoke tests, the build, and typecheck passed; the smoke suite also passed alongside the existing CLI symlink invocation/build test. The wrong-version mutation was correctly detected. However, the heartbeat no-op and JSON timeline error-exit mutations both left all seven tests green, as detailed inline. A delayed child also demonstrated that Vitest's timeout does not interrupt spawnSync.

Local validation was on macOS; I did not rerun the full regression suite.

abhinav-phi:
Fixed the three findings in 089df37: a 10-second subprocess timeout with diagnostics, timeline JSON exit-status assertion, and HEARTBEAT_OK output assertion. Each regression mutation now fails the test. Final-head CI is green at 5f15d3d. @theDakshJaitly please re-review when you can.

abhinav-phi:
hi the review fixes are pushed in 089df37

every spawned child process is bounded with an explicit timeout and killed on expiry

failures include error signal and stderr in the assertion messages

exit status is asserted before json parsing and heartbeat output must contain the ok marker

all ci checks are green

could you take another look when you get time

theDakshJaitly (approved):
Re-reviewed the final head: all three requested changes are resolved, with no remaining findings. Verified that a hanging child is killed after 10 seconds with useful diagnostics, a nonzero timeline JSON exit fails its status assertion, and a no-output heartbeat fails its output assertion. The wrong-version mutation is also detected. The seven smoke tests pass, and eight tests pass when combined with the existing CLI build/symlink check. Build, typecheck, and final-head CI are green.

Issue #42, opened 2026-05-14
Decide and document zero-day heartbeat threshold behavior


theDakshJaitly:
Follow-up from Copilot review on PR #39.

## Problem
heartbeat.staleDays, memoryCleanupDays, and dailyMemoryRetentionDays currently use positive-number validation, so 0 is rejected and the default is used instead.

That means config cannot express: consider stale as soon as older than today.

## Decision needed
Choose one of these behaviors:

1. Allow 0 for day-based heartbeat thresholds while still rejecting negatives.
2. Keep requiring positive values, but document that 0 is intentionally unsupported.

## Acceptance criteria
- The intended behavior is documented in README/config docs.
- Tests reflect the chosen behavior.
- If 0 is allowed, config parsing accepts >= 0 for day-based heartbeat thresholds only.

Source: https://github.com/theDakshJaitly/mex/pull/39#discussion_r3238408317

abhinav-phi:
Hi @theDakshJaitly,

I'd like to work on this issue. If it's still available, could you please assign it to me?

Thanks!

Issue #41, opened 2026-05-14
Show heartbeat hint when legacy scaffolds have no last_updated fields


theDakshJaitly:
Follow-up from Copilot review on PR #39.

## Problem
Heartbeat staleness is opt-in via last_updated frontmatter. That is good for v0.2 compatibility, but legacy scaffolds with no last_updated fields can make mex heartbeat report a clean status even though no files are actually eligible for staleness checks.

## Suggested fix
When heartbeat scans zero scaffold files with parseable last_updated, print a short hint explaining that staleness checks are inactive until files opt in with last_updated.

Example intent:
No scaffold files include last_updated; staleness checks are currently skipped.

## Acceptance criteria
- mex heartbeat surfaces a hint when zero files participate in staleness checks.
- The hint does not turn an otherwise healthy heartbeat into an error.
- Existing v0.2 scaffolds remain compatible.
- TUI/doctor behavior is updated if they surface heartbeat status.

Source: https://github.com/theDakshJaitly/mex/pull/39#discussion_r3238408361

abhinav-phi:
Hi @theDakshJaitly,

I'd like to work on this issue. If it's still available, could you please assign it to me?

Thanks!

PR #39, opened 2026-05-14
Release v0.3 operational memory

theDakshJaitly:
## Summary

Prepares the mex v0.3 operational-memory release on top of the stable v0.2 architecture.

This includes:
- agent-memory setup mode for persistent-agent / homelab / OpenClaw-style workspaces
- heartbeat checks and `mex watch --interval`
- append-only event log with `mex log` and `mex timeline`
- `mex doctor`, shell completions, and improved check output
- Ink TUI dashboard shipped through bare `mex` and `mex tui`
- npm release prep for `promexeus@0.3.5`
- copy-ready GitHub release notes in `RELEASE_NOTES.md`

## Release Notes

Full release notes are in `RELEASE_NOTES.md` and cover:
- agent memory layer
- heartbeat / scheduled checks
- TUI dashboard
- OpenClaw and persistent-agent usage
- upgrade instructions
- compatibility and deferred architecture work

## Verification

- `npm run typecheck`
- `npm test` — 138 passing
- `npm run build`
- `node dist/cli.js --version` → `0.3.5`
- `npm --cache /private/tmp/mex-npm-cache pack --dry-run` → `promexeus-0.3.5.tgz`
- `git diff --check`
- manual TTY smoke for `node dist/cli.js tui`

PR #181, opened 2026-09-08
fix(heartbeat): Copilot-review follow-ups — inactive-staleness hint, zero-day thresholds, symlink dedup

### LINKED ISSUE #40: Handle symlinked scaffold files safely in heartbeat and drift scans
Follow-up from Copilot review on PR #39.

## Problem
scaffoldHeartbeatFiles and the drift scanner use globbing with follow: true. This supports symlinked scaffold content, but it may also create duplicate file entries or unsafe traversal behavior if users have symlink loops inside .mex/.

## Suggested fix
Confirm the intended symlink behavior and either:
- de-duplicate scanned files by real path, or
- stop following symlinks where unsafe, or
- document the behavior clearly for users upgrading from v0.2.

## Acceptance criteria
- Symlinked scaffold files do not produce duplicate heartbeat/drift entries.
- Cyclic symlinks do not cause runaway scanning.
- Behavior is covered by a focused test or documented as unsupported.

Source: https://github.com/theDakshJaitly/mex/pull/39#discussion_r3238408343

abhinav-phi:
Resolves #40, #41, and #42 — the three follow-ups from the Copilot review on PR #39, all touching the heartbeat/drift scan surface.

## #41 — surface inactive staleness checks
`checkHeartbeat` returns `filesWithoutLastUpdated` (additive, absent once any file opts in) when every scanned scaffold file lacks a parseable `last_updated`. `mex heartbeat` prints the skip-hint with opt-in instructions; `mex doctor`'s Heartbeat line notes the inactive checks. Exit codes and `ok` untouched.

## #42 — accept zero for day-based thresholds
`staleDays`, `memoryCleanupDays`, `dailyMemoryRetentionDays` now accept `0` ("stale as soon as older than today"); negatives and garbage still fall back to the default. `watch.intervalMinutes` keeps positive validation. Documented on `HeartbeatConfig`.

## #40 — deduplicate symlinked scaffold files
Both `scaffoldHeartbeatFiles` and `findScaffoldFiles` deduplicate by realpath, so one file reachable through two patterns or a symlinked directory is scanned once. Glob bounds symlink loops; the duplicate-entry risk was the real defect.

## Tests
10 heartbeat tests (incl. symlink-dedup with a Windows no-privilege skip guard) + updated config parsing tests; `npm run typecheck` green.

### DIFF
--- a/src/drift/index.ts
+++ b/src/drift/index.ts
@@ -1,4 +1,4 @@
-import { readFileSync } from "node:fs";
+import { readFileSync, realpathSync } from "node:fs";
 import { resolve, relative } from "node:path";
 import { globSync } from "glob";
 import type { MexConfig, DriftReport, DriftIssue, Claim } from "../types.js";
@@ -419,15 +419,28 @@ export function findScaffoldFiles(
 ): string[] {
   const files: string[] = [];
 
-  // Search inside scaffold root (handles both .mex/ and root layouts)
+  // Search inside scaffold root (handles both .mex/ and root layouts).
+  // follow: true supports symlinked scaffold content; deduplicating by real
+  // path keeps one file reached through two links a single scan entry, and
+  // glob bounds symlink loops so runaway scans stay off the table (#40).
+  const seenReal = new Set<string>();
   for (const pattern of patterns) {
-    const matches = globSync(pattern, {
+    for (const match of globSync(pattern, {
       cwd: scaffoldRoot,
       absolute: true,
       follow: true,
       ignore: ["node_modules/**"],
-    });
-    files.push(...matches);
+    })) {
+      let real: string;
+      try {
+        real = realpathSync(match);
+      } catch {
+        real = match;
+      }
+      if (seenReal.has(real)) continue;
+      seenReal.add(real);
+      files.push(match);
+    }
   }
 
   // Also check project root for tool config files (CLAUDE.md, etc.)
--- a/src/heartbeat.ts
+++ b/src/heartbeat.ts
@@ -12,6 +12,15 @@ export interface HeartbeatResult {
   staleFiles: Array<{ file: string; days: number }>;
   memoryCleanupDue: boolean;
   oldDailyMemoryFiles: string[];
+  /**
+   * Additive: number of scaffold files the heartbeat scanned that carry no
+   * parseable `last_updated` at all. When every scanned file lacks one,
+   * staleness checks are silently inert — the CLI surfaces a hint instead of
+   * reporting a healthy-looking sweep that checked nothing. Absent (0) for
+   * scaffolds where at least one file opts in, so existing JSON consumers
+   * see no change.
+   */
+  filesWithoutLastUpdated?: number;
 }
 
 export interface HeartbeatOpts {
@@ -63,9 +72,13 @@ export function checkHeartbeat(
   const memoryCleanupDays = config.heartbeat?.memoryCleanupDays ?? DEFAULT_MEMORY_CLEANUP_DAYS;
   const dailyRetentionDays = config.heartbeat?.dailyMemoryRetentionDays ?? DEFAULT_DAILY_MEMORY_RETENTION_DAYS;
 
+  let scanned = 0;
+  let withLastUpdated = 0;
   const staleFiles = scaffoldHeartbeatFiles(config.scaffoldRoot, opts.scaffoldPatterns)
     .map((file) => {
       const fm = parseFrontmatter(file);
+      scanned++;
+      if (typeof fm?.last_updated === "string") withLastUpdated++;
       const days = daysSinceFrontmatterDate(
         typeof fm?.last_updated === "string" ? fm.last_updated : undefined,
         now,
@@ -84,6 +97,7 @@ export function checkHeartbeat(
     staleFiles,
     memoryCleanupDue,
     oldDailyMemoryFiles,
+    ...(scanned > 0 && withLastUpdated === 0 ? { filesWithoutLastUpdated: scanned } : {}),
   };
 }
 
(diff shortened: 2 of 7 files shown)


src/heartbeat.ts
```diff
@@ -63,9 +72,13 @@ export function checkHeartbeat(
   const memoryCleanupDays = config.heartbeat?.memoryCleanupDays ?? DEFAULT_MEMORY_CLEANUP_DAYS;
   const dailyRetentionDays = config.heartbeat?.dailyMemoryRetentionDays ?? DEFAULT_DAILY_MEMORY_RETENTION_DAYS;
 
+  let scanned = 0;
+  let withLastUpdated = 0;
   const staleFiles = scaffoldHeartbeatFiles(config.scaffoldRoot, opts.scaffoldPatterns)
     .map((file) => {
       const fm = parseFrontmatter(file);
+      scanned++;
+      if (typeof fm?.last_updated === "string") withLastUpdated++;
```
theDakshJaitly: [P2] Count parseable dates, not every string

`withLastUpdated` is incremented before the date is validated. A scaffold containing only `last_updated: ""`, `"not-a-date"`, or `"[YYYY-MM-DD]"` gets `ok: true` with no inactive hint, even though `daysSinceFrontmatterDate` rejects every value and no files participate. I reproduced all three cases. Please increment the participation count only for dates accepted by the staleness check, and cover invalid strings as well as missing fields so this satisfies #41's parseable-date requirement.

src/drift/index.ts
```diff
@@ -419,15 +419,28 @@ export function findScaffoldFiles(
...
+    })) {
+      let real: string;
+      try {
+        real = realpathSync(match);
+      } catch {
+        real = match;
+      }
+      if (seenReal.has(real)) continue;
+      seenReal.add(real);
+      files.push(match);
```
theDakshJaitly, line 434: [P2] Deduplicate root tool-file aliases too

The new realpath deduplication only covers this scaffold loop. The following loop appends root tool files directly, and the final Set compares path spellings. With `CLAUDE.md` symlinked to `.mex/AGENTS.md`, `findScaffoldFiles` still returns both paths for the same physical file, so it is scanned twice. Please apply the same realpath deduplication to both discovery phases and add a root-tool alias regression case.

test/heartbeat.test.ts
```diff
@@ -46,8 +46,89 @@ describe("heartbeat", () => {
...
+    const root = mkdtempSync(join(tmpdir(), "mex-heartbeat-symlink-"));
+    try {
+      mkdirSync(join(root, ".mex/context"), { recursive: true });
+      mkdirSync(join(root, ".mex/extra"), { recursive: true });
+      const real = join(root, ".mex/context/architecture.md");
+      writeFileSync(real, frontmatter("architecture", "2026-05-01"));
+      // Same content reachable through a second pattern's directory.
+      let linked = true;
+      try {
+        symlinkSync(real, join(root, ".mex/extra/architecture.md"));
```
theDakshJaitly: [P2] Make the symlink test actually exercise deduplication

`symlinkSync` is not imported, so this throws a ReferenceError that the broad catch treats as a symlink-permission failure; `if (!linked) return` then skips the assertion on every platform. Rethrowing in a diagnostic copy confirmed this. There is a second false-pass issue: `.mex/extra/` is outside the default heartbeat glob patterns. After supplying the import, this test still passes against main without deduplication. Please import the function, skip only known permission errors, and put the alias in a scanned directory such as `patterns/` or explicitly pass patterns that visit both paths. Add drift-side coverage too.

src/heartbeat.ts
```diff
@@ -12,6 +12,15 @@ export interface HeartbeatResult {
...
   oldDailyMemoryFiles: string[];
+  /**
+   * Additive: number of scaffold files the heartbeat scanned that carry no
+   * parseable `last_updated` at all. When every scanned file lacks one,
+   * staleness checks are silently inert — the CLI surfaces a hint instead of
+   * reporting a healthy-looking sweep that checked nothing. Absent (0) for
+   * scaffolds where at least one file opts in, so existing JSON consumers
+   * see no change.
+   */
+  filesWithoutLastUpdated?: number;
```
theDakshJaitly, line 23: [P2] Show inactive staleness in the TUI

This new status is consumed by the CLI output but never by `Summary`, `HeartbeatPanel`, or `DoctorPanel` in `src/tui.ts`. For a legacy scaffold with `filesWithoutLastUpdated: 1`, I reproduced `HeartbeatPanel` rendering `HEARTBEAT_OK · scaffold is fresh`, with no indication that staleness checks are inactive. Please propagate the inactive/opt-in hint to the TUI views while preserving healthy exit semantics, as required by #41.

theDakshJaitly (changes requested):
The zero-day threshold behavior works, including all three persisted heartbeat settings while zero watch intervals remain rejected. Requesting changes for the four findings inline: count only eligible dates for the inactive hint, include root tool-file aliases in deduplication, make the symlink test exercise the fix, and surface inactive staleness in the TUI.

I tested the exact PR diff on main at 9648c9d. All 97 existing focused tests, npm run typecheck, npm run build, and git diff --check passed. Targeted probes reproduced the invalid-date, root-alias, and TUI gaps. Diagnostic copies of the symlink test confirmed the swallowed ReferenceError and showed that adding the import alone still lets it pass against the pre-fix implementation.

Local validation was on macOS; the full regression suite and Windows tests were not rerun locally.

src/heartbeat.ts
abhinav-phi: Fixed in b687bd0: participation now increments only when daysSinceFrontmatterDate returns a non-null result. The empty string, not-a-date, and placeholder cases failed before the fix and pass now.

src/drift/index.ts
abhinav-phi: Fixed in b687bd0. Both discovery phases now feed the same realpath deduplication. The new root-alias tests fail on the previous code (4 instead of 2 files, and 2 instead of 1) and pass on Linux with real symlinks.

test/heartbeat.test.ts
abhinav-phi: Fixed in b687bd0: imported symlinkSync, moved the alias to patterns/, and limited skipping to Windows EPERM/EACCES. Added drift-side cases too. Linux runs have zero skips; disabling heartbeat dedup makes the corrected test fail with two stale entries.

src/heartbeat.ts
abhinav-phi: Fixed in b687bd0 in src/tui.ts: Summary, HeartbeatPanel, and DoctorPanel now show the inactive/opt-in hint, including memory-attention cases. Healthy semantics are preserved. test/tui.test.ts passes 19/19; final-head CI is green. @theDakshJaitly please take another look.

abhinav-phi:
hi i pushed fixes for all four review threads in b687bd0

files with invalid or unparseable dates no longer count as staleness participants and heartbeat shows the inactive hint with the opt in guidance

scaffold files dedupe by real path including the root tool files so aliases count exactly once

the symlink test uses real symlinks with explicit windows permission skips

tui and doctor show the inactive staleness state without changing healthy exit semantics

all ci checks are green

could you take another look when you get time

theDakshJaitly (approved):
Re-reviewed the final head: all four requested changes are resolved, with no remaining findings. Invalid dates now trigger the inactive-check hint; realpath deduplication covers scaffold and root tool aliases; the corrected symlink tests exercise real aliases; and Summary, HeartbeatPanel, and DoctorPanel display the inactive state and opt-in guidance. Verified 128 focused tests, including the built CLI smoke suite, plus build and typecheck. Removing each fix causes the corresponding regression tests to fail. Final-head CI is green.

PR #182, opened 2026-09-09
feat(cli): add timeline --format md

### LINKED ISSUE #55: Add `mex timeline --format md`
## Context

`mex timeline` currently renders to the terminal. Add a `--format md` option that emits Markdown, suitable for piping into reports / standup notes.

## Scope

- Add a `--format` option to the `timeline` command (default: current terminal output; `md`: Markdown).
- Markdown output should be valid and readable (a table or a clean list of events).

## Where

- `timeline` command in `src/cli.ts` and wherever timeline rendering lives (grep for `timeline`).

## Acceptance criteria

- [ ] `mex timeline --format md` emits valid Markdown.
- [ ] Default output is unchanged.
- [ ] A test covers the `md` output shape.

abhinav-phi:
Resolves #55.

## What

`mex timeline --format md` renders the (already-filterable — `--since`, `--type`, `--limit` all compose) timeline as a Markdown table:

```markdown
| Date | Type | Event | Files |
|---|---|---|---|
| 2026-09-09 | decision | chose the bounded resolver | `ROUTER.md` |
```

- Messages are escaped (pipes → `\|`, line breaks → spaces) so a message cannot break the table; file paths render as inline code.
- An empty log emits `_No events found._` — valid Markdown, still readable when pasted.
- `--json` takes precedence (unchanged); the **default terminal output is byte-identical** when `--format` is absent.

## Tests

Three new cases in `test/events.test.ts`: table shape + escaping with a pipe-bearing message and a file column, empty-log placeholder, and default output unchanged. `npm run typecheck` green.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -897,6 +897,7 @@ program
   .command("timeline")
   .description("Read bounded recent project notes (latest 8 MiB / 10,000 log lines)")
   .option("--json", "Output events as JSON")
+  .option("--format <format>", "Output format: md for a Markdown table (piped into reports)")
   .option("--since <date>", "Filter from YYYY-MM-DD or relative Nd, e.g. 30d")
   .option("--type <type>", "Filter by event type: decision, note, risk, todo")
   .option("--query <text>", "Case-insensitive literal text in the message (max 256 UTF-8 bytes)")
--- a/src/events.ts
+++ b/src/events.ts
@@ -58,6 +58,8 @@ export interface LogOpts {
 
 export interface TimelineOpts {
   json?: boolean;
+  /** `"md"` renders a Markdown table instead of the terminal output (#55). */
+  format?: string;
   since?: string;
   kind?: string;
   limit?: number;
@@ -143,6 +145,11 @@ export async function runTimeline(config: MexConfig, opts: TimelineOpts = {}): P
     return;
   }
 
+  if (opts.format === "md") {
+    printTimelineMarkdown(selected, truncated, sourceTruncated);
+    return;
+  }
+
   if (selected.length === 0 && !truncated) {
     console.log(chalk.dim("No events found."));
   }
@@ -153,6 +160,32 @@ export async function runTimeline(config: MexConfig, opts: TimelineOpts = {}): P
   }
   if (truncated) console.log(chalk.dim("Some matching events were omitted by the entry or output limit; narrow the filters."));
   if (sourceTruncated) console.log(chalk.dim("Searched only the latest 8 MiB / 10,000 non-empty log lines; older history was not scanned."));
+}
+
+/**
+ * Markdown rendering for `timeline --format md` — valid inside reports and
+ * standup notes. Pipes and line breaks are escaped so a message cannot break
+ * the table; omission notes mirror the terminal output in italics. The default
+ * terminal rendering is untouched (#55).
+ */
+function printTimelineMarkdown(
+  selected: EventEntry[],
+  truncated: boolean,
+  sourceTruncated: boolean,
+): void {
+  if (selected.length === 0) {
+    console.log("_No events found._");
+    return;
+  }
+  console.log("| Date | Type | Event | Files |");
+  console.log("|---|---|---|---|");
+  for (const e of selected) {
+    const message = e.message.replace(/\|/g, "\\|").replace(/\r?\n/g, " ");
+    const files = e.files.length ? e.files.map((f) => `\`${f}\``).join(", ") : "—";
+    console.log(`| ${e.timestamp.slice(0, 10)} | ${e.kind} | ${message} | ${files} |`);
+  }
+  if (truncated) console.log("_Some matching events were omitted by the entry or output limit; narrow the filters._");
+  if (sourceTruncated) console.log("_Searched only the latest 8 MiB / 10,000 non-empty log lines; older history was not scanned._");
 }
 
 export function readEvents(config: MexConfig): EventEntry[] {
--- a/test/events.test.ts
+++ b/test/events.test.ts
@@ -188,4 +188,32 @@ describe("events", () => {
     await runTimeline(config, { json: true });
     expect(spy.mock.calls.at(-1)?.[0]).toContain('"events"');
   });
+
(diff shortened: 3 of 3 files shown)


src/cli.ts
```diff
@@ -897,6 +897,7 @@ program
   .command("timeline")
   .description("Read bounded recent project notes (latest 8 MiB / 10,000 log lines)")
   .option("--json", "Output events as JSON")
+  .option("--format <format>", "Output format: md for a Markdown table (piped into reports)")
```
theDakshJaitly, line 900: **[P1] Forward the format option to the timeline handler**

The action below builds a new options object for `runTimeline` but never includes `format: opts.format`. As a result, the built CLI accepts `mex timeline --format md` and exits successfully while still printing ordinary terminal text; the new renderer is unreachable through the advertised command. Please forward the option and add a test invoking the actual CLI, since the new tests call `runTimeline` directly and bypass this dispatch.

src/events.ts
```diff
@@ -155,6 +162,32 @@ export async function runTimeline(config: MexConfig, opts: TimelineOpts = {}): P
...
+ * terminal rendering is untouched (#55).
+ */
+function printTimelineMarkdown(
+  selected: EventEntry[],
+  truncated: boolean,
+  sourceTruncated: boolean,
+): void {
+  if (selected.length === 0) {
+    console.log("_No events found._");
+    return;
```
theDakshJaitly: **[P2] Preserve omission notices when the selected result is empty**

This early return skips both `truncated` and `sourceTruncated` notices. A matching event with a 70,000-character message is excluded by the output budget and produces only `_No events found._`; likewise, a query matching only an older entry outside the retained 10,000-line window hides the incomplete source scan. Please emit the applicable omission notices even when no rows are selected, and distinguish an omitted matching event from a genuinely empty result.

src/events.ts
```diff
@@ -155,6 +162,32 @@ export async function runTimeline(config: MexConfig, opts: TimelineOpts = {}): P
...
+  }
+  console.log("| Date | Type | Event | Files |");
+  console.log("|---|---|---|---|");
+  for (const e of selected) {
+    const message = e.message.replace(/\|/g, "\\|").replace(/\r?\n/g, " ");
+    const files = e.files.length ? e.files.map((f) => `\`${f}\``).join(", ") : "—";
```
theDakshJaitly: **[P2] Encode messages and file paths as Markdown table cells**

File paths are inserted verbatim inside single backticks, so accepted paths containing a pipe (such as `src/a|b.ts`), newline, or backtick break the table or code span. The message escaping also fails when a pipe already has a backslash before it, and leaves bare carriage returns intact. These cases reproduce with the renderer and a Markdown table parser. Please use table-aware encoding for both messages and paths, handling backslashes, CR/LF, and code-span delimiters, and test the rendered cell contents.

src/events.ts
```diff
@@ -155,6 +162,32 @@ export async function runTimeline(config: MexConfig, opts: TimelineOpts = {}): P
...
+    console.log(`| ${e.timestamp.slice(0, 10)} | ${e.kind} | ${message} | ${files} |`);
```
theDakshJaitly: **[P2] Apply the byte limit to the rendered Markdown**

The shared selection budget measures JSON serialization, but Markdown escaping can make the emitted rows substantially larger. A single accepted message containing 40,000 pipe characters produces 80,081 UTF-8 bytes here, exceeding the 65,536-byte output limit. Please budget the actual escaped rows together with the table header and omission notices before emitting them, and report when a complete row cannot fit.

theDakshJaitly (changes requested):
Requesting changes for the four inline findings. The CLI currently drops the new format option, so `mex timeline --format md` still prints ordinary terminal output. The Markdown renderer also needs to preserve omission notices, correctly encode table cells, and enforce the output limit after rendering.

Validation: applied the exact PR diff to main in an isolated checkout; 53 existing events/public API tests, typecheck, build, and diff checks passed. Targeted probes reproduced the four issues. CLI forwarding was also checked on the PR head.

src/cli.ts
abhinav-phi: Fixed in 7de394c: the CLI now forwards format. The built-CLI regression checks Markdown dispatch, default output, and JSON precedence. Child processes and the outer test are bounded in a03a4ba/f08fbd1; the exact test passes locally. Final-head CI is green: https://github.com/mex-memory/mex/actions/runs/35165138729

src/events.ts
abhinav-phi: Fixed in 7de394c. Empty selection now retains both omission notices and distinguishes budget-omitted matches from no matches. Added regressions for oversized entries and matches outside the retained history window.

src/events.ts
abhinav-phi: Fixed in 7de394c. Messages and paths handle pipes, backslashes, CR/LF, and backtick runs. Regression tests parse the output with GFM and verify four cells and round-tripped content.

src/events.ts
abhinav-phi: Fixed in 7de394c. Markdown rows are budgeted after escaping, including UTF-8 bytes, headers, newlines, and notices; oversized rows are omitted whole. All 48 events tests pass, including the byte-boundary cases. Final-head CI is green; @theDakshJaitly could you re-review?

abhinav-phi:
hi all four review points are fixed and pushed

the cli forwards the format option so timeline with format md actually renders markdown

empty selections keep both omission notices and budget omitted results are separate from no matches

messages and file paths are encoded as proper table cells handling pipes backslashes newlines and backtick runs

the byte limit applies to the rendered markdown and oversized rows are dropped whole

the events suite passes fully and final head ci is green

could you re review when you get time

theDakshJaitly (approved):
Re-reviewed the updated commits against all four requested changes. CLI format forwarding, omission notices for empty selections, GFM cell encoding, and the rendered 64 KiB output bound are now fixed. No remaining findings.

Verified the complete PR patch on current main: 91 focused tests passed, including the original review probes, the built-CLI output regression, public API coverage, and CLI smoke tests. Build, workspace typecheck, and diff checks passed; default and JSON output remain unchanged. CI is green on this head.

PR #183, opened 2026-09-09
feat(cli): add mex export (bundle scaffold to a single Markdown file)

### LINKED ISSUE #56: Add `mex export` (bundle scaffold to a single Markdown file)
## Context

Add a `mex export` command that bundles the whole scaffold into a single Markdown file — for pasting into tools that don't read files well.

## Scope

- New `mex export` command that concatenates the scaffold files (ROUTER.md, context/*, patterns/*, etc.) into one Markdown document, with a clear section header per source file.
- Output to stdout by default; optionally to a path.

## Where

- New command in `src/cli.ts`.
- Reuse scaffold-file discovery — see `DEFAULT_SCAFFOLD_PATTERNS` and `findScaffoldFiles` in `src/drift/index.ts`.

## Acceptance criteria

- [ ] `mex export` produces a single Markdown file containing all scaffold files, each under a header.
- [ ] Works with stdout and (if implemented) a target path.
- [ ] A test covers the bundled output.

abhinav-phi:
Resolves #56.

## What

`mex export` concatenates the whole scaffold into one Markdown document, for pasting into tools that don't read files well:

- **Same discovery as `mex check`** — reuses `findScaffoldFiles` + `DEFAULT_SCAFFOLD_PATTERNS`, so the export is exactly what the drift scanner sees, never a divergent file list.
- Each file lands under a `## <scaffold-relative-path>` header, with content trimmed of trailing whitespace so the document stays clean.
- Deterministic order (sorted by path), so repeated exports diff cleanly.
- `mex export` → stdout; `mex export --out exports/scaffold.md` → file (parent directories created) plus a one-line count report.
- An empty scaffold fails with `No scaffold files found. Run: mex setup`.

## Tests

Three cases in `test/export.test.ts`: full bundle (headers, content, deterministic order), `--out` write + count report, and missing-scaffold guidance. `npm run typecheck` green.

### DIFF
--- a/src/export.ts
+++ b/src/export.ts
@@ -1,0 +1,43 @@
+import { mkdirSync, readFileSync, writeFileSync } from "node:fs";
+import { dirname, relative, resolve } from "node:path";
+import { DEFAULT_SCAFFOLD_PATTERNS, findScaffoldFiles } from "./drift/index.js";
+import { toPosix } from "./paths.js";
+import type { MexConfig } from "./types.js";
+
+export interface ExportOpts {
+  /** Write the bundle to this file instead of stdout. */
+  out?: string;
+}
+
+/**
+ * Bundle the whole scaffold into one Markdown document (#56).
+ *
+ * Section headers name the source file so a pasted copy stays navigable, and
+ * files are emitted in a deterministic order (sorted by path). Reuses the
+ * drift scanner's own file discovery, so what gets exported is exactly what
+ * `mex check` scans — nothing drifts between the two.
+ */
+export async function runExport(config: MexConfig, opts: ExportOpts = {}): Promise<void> {
+  const files = findScaffoldFiles(config.projectRoot, config.scaffoldRoot, DEFAULT_SCAFFOLD_PATTERNS)
+    .sort((left, right) => (left < right ? -1 : left > right ? 1 : 0));
+
+  if (files.length === 0) {
+    throw new Error("No scaffold files found. Run: mex setup");
+  }
+
+  const bundle: string[] = ["# mex scaffold export", ""];
+  for (const file of files) {
+    const relativePath = toPosix(relative(config.scaffoldRoot, file));
+    bundle.push(`## ${relativePath}`, "", readFileSync(file, "utf-8").trimEnd(), "");
+  }
+  const document = bundle.join("\n");
+
+  if (opts.out) {
+    const target = resolve(config.projectRoot, opts.out);
+    mkdirSync(dirname(target), { recursive: true });
+    writeFileSync(target, document, "utf-8");
+    console.log(`Wrote ${files.length} scaffold file(s) to ${opts.out}`);
+    return;
+  }
+  process.stdout.write(document);
+}
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -920,6 +920,21 @@ program
   });
 
 program
+  .command("export")
+  .description("Bundle the whole scaffold into a single Markdown document")
+  .option("--out <path>", "Write to a file instead of stdout")
+  .action(async (opts) => {
+    try {
+      const config = loadConfig();
+      const { runExport } = await import("./export.js");
+      await runExport(config, opts);
+    } catch (err) {
+      console.error((err as Error).message);
+      process.exit(1);
+    }
+  });
+
+program
   .command("timeline")
   .description("Show recent mex event log entries")
   .option("--json", "Output events as JSON")
--- a/test/export.test.ts
+++ b/test/export.test.ts
@@ -1,0 +1,59 @@
+import { describe, it, expect, beforeEach, afterEach, vi } from "vitest";
+import { mkdtempSync, mkdirSync, readFileSync, rmSync, writeFileSync } from "node:fs";
+import { join } from "node:path";
+import { tmpdir } from "node:os";
+import { runExport } from "../src/export.js";
+import type { MexConfig } from "../src/types.js";
+
+let tmpDir: string;
+let config: MexConfig;
+let stdoutSpy: ReturnType<typeof vi.spyOn>;
+
+beforeEach(() => {
+  tmpDir = mkdtempSync(join(tmpdir(), "mex-export-"));
+  mkdirSync(join(tmpDir, ".mex/context"), { recursive: true });
+  mkdirSync(join(tmpDir, ".mex/patterns"), { recursive: true });
+  writeFileSync(join(tmpDir, ".mex/ROUTER.md"), "# Router\n\nEntry point.\n");
+  writeFileSync(join(tmpDir, ".mex/context/stack.md"), "# Stack\n\nNode 22.\n");
+  writeFileSync(join(tmpDir, ".mex/patterns/retry.md"), "# Retry\n");
+  config = { projectRoot: tmpDir, scaffoldRoot: join(tmpDir, ".mex"), aiTools: [] };
+  stdoutSpy = vi.spyOn(process.stdout, "write").mockImplementation(() => true);
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+  rmSync(tmpDir, { recursive: true, force: true });
+});
+
+describe("mex export (#56)", () => {
+  it("bundles every scaffold file under a header per source file", async () => {
+    await runExport(config, {});
+    const document = stdoutSpy.mock.calls.map((call) => String(call[0])).join("");
+
+    expect(document).toContain("# mex scaffold export");
+    expect(document).toContain("## ROUTER.md");
+    expect(document).toContain("## context/stack.md");
+    expect(document).toContain("## patterns/retry.md");
+    // Content survives intact under its own header.
+    expect(document).toContain("Entry point.");
+    expect(document).toContain("Node 22.");
+    // Deterministic order: sorted by path.
+    expect(document.indexOf("## ROUTER.md")).toBeLessThan(document.indexOf("## context/stack.md"));
+    expect(document.indexOf("## context/stack.md")).toBeLessThan(document.indexOf("## patterns/retry.md"));
+  });
+
+  it("writes the same bundle to --out and reports the count", async () => {
+    const logSpy = vi.spyOn(console, "log").mockImplementation(() => {});
+    await runExport(config, { out: "exports/scaffold.md" });
+    expect(stdoutSpy.mock.calls.map((call) => String(call[0])).join("")).toBe("");
+    const written = readFileSync(join(tmpDir, "exports/scaffold.md"), "utf-8");
+    expect(written).toContain("## ROUTER.md");
+    expect(logSpy.mock.calls.map((call) => String(call[0])).join(""))
+      .toContain("Wrote 3 scaffold file(s) to exports/scaffold.md");
+  });
+
+  it("fails with guidance when the scaffold is missing", async () => {
+    rmSync(join(tmpDir, ".mex"), { recursive: true, force: true });
+    await expect(runExport(config, {})).rejects.toThrow("No scaffold files found. Run: mex setup");
+  });
+});
--- a/test/wiki-architecture.test.ts
+++ b/test/wiki-architecture.test.ts
@@ -761,6 +761,7 @@ describe("no unscoped scaffold writes", () => {
       "src/config.ts": "writes config.json",
       "src/global-config.ts": "writes the global config and telemetry id",
       "src/events.ts": "appends to events/decisions.jsonl",
+      "src/export.ts": "writes one export bundle to a user-specified path — a brand-new file, never scaffold bytes",
       "src/pattern/index.ts": "creates a new pattern file from a template",
       "src/setup/anchor.ts": "edits root tool configs only, never .mex/, and only inside its own markers",
       "src/setup/ignore.ts": "creates or appends only the setup-managed .mex/.gitignore rules",


abhinav-phi:
hi this adds the export command that bundles the scaffold into a single markdown file

the output is bounded and paths stay inside the scaffold

happy to adjust the format if you want a different shape

src/export.ts
```diff
@@ -0,0 +1,43 @@
...
+  for (const file of files) {
+    const relativePath = toPosix(relative(config.scaffoldRoot, file));
+    bundle.push(`## ${relativePath}`, "", readFileSync(file, "utf-8").trimEnd(), "");
+  }
+  const document = bundle.join("\n");
+
+  if (opts.out) {
+    const target = resolve(config.projectRoot, opts.out);
+    mkdirSync(dirname(target), { recursive: true });
+    writeFileSync(target, document, "utf-8");
```
theDakshJaitly, line 148: [P1] Refuse export destinations that overwrite project state

`writeFileSync` truncates any existing target, including the files being exported. In the built CLI, `mex export --out .mex/ROUTER.md` replaced the Router with the bundle, and `--out .mex/config.json` replaced configuration JSON with Markdown; both exited 0. An `exports/scaffold.md` symlink to the Router also overwrote it. This contradicts the new architecture exemption's claim that this writer creates a brand-new file and never writes scaffold bytes. Validate the destination and its aliases before writing, preserve existing scaffold/configuration state, and enforce the intended new-file behavior safely. Cover these rejected destinations with tests asserting that the original bytes survive. Also prevent output from becoming an input: repeated exports to `.mex/context/bundle.md` included the prior bundle and duplicated the scaffold.

src/export.ts
```diff
@@ -0,0 +1,43 @@
...
+  if (files.length === 0) {
+    throw new Error("No scaffold files found. Run: mex setup");
+  }
+
+  const bundle: string[] = ["# mex scaffold export", ""];
+  for (const file of files) {
+    const relativePath = toPosix(relative(config.scaffoldRoot, file));
+    bundle.push(`## ${relativePath}`, "", readFileSync(file, "utf-8").trimEnd(), "");
+  }
+  const document = bundle.join("\n");
```
theDakshJaitly, line 143: [P2] Bound scaffold reads before retaining the full export

`findScaffoldFiles` provides discovery and deduplication, not file-count or byte limits. This loop reads every file into memory, retains the whole corpus, then allocates the joined document before either output mode can write anything. Large scaffold content can therefore terminate Node instead of producing a controlled error. In a deliberately constrained 64 MiB V8-heap diagnostic, the small fixture passed but two 40 MiB source files caused a fatal heap error during `ReadFileUtf8`; this is not a claim about Node's default heap threshold. MEX's bounded-input/output and retained-state rules require explicit file/count/aggregate limits checked before allocation, with a clear refusal and boundary tests. If larger exports are supported, use a bounded streaming design instead of retaining the full document.

theDakshJaitly (changes requested):
The normal export works, but the destination needs protection against overwriting project state, and scaffold reads need explicit resource bounds. Both findings are inline.

Validation on current main with this PR applied: 59 existing tests, production build, workspace typecheck, and diff checks passed. Additional built-CLI probes reproduced the overwrite and resource-limit failures. Stdout and file output matched for the normal fixture.

src/export.ts
abhinav-phi: Addressed in 2abae42, with the destination validated before anything is written.

How it works now: `--out` is resolved (including through symlinks, via best-effort realpath that also resolves symlinked parent directories) and compared against every scaffold file, the project config (`.mex/config.json`, which discovery never lists, so it gets an explicit guard), and their realpath aliases. A collision throws before any write, so the original bytes always survive. An existing target carrying the bundle marker (`# mex scaffold export`) is treated as a previous output rather than project state: it is excluded from its own inputs and may be overwritten, which also fixes the repeat-export duplication (second run is now byte-identical to the first).

Tests in `test/export.test.ts` prove original bytes survive for all three reported cases: `--out .mex/ROUTER.md`, `--out .mex/config.json`, and a symlink alias. One environment note: the file-symlink fixture itself needs symlink privilege, so it errors with EPERM on my Windows box; the same alias-detection path is additionally covered by a directory-junction variant that runs everywhere, and the symlink case will run on Linux CI. The architecture test (pinned writers, guard table) still passes since no write call sites were added.

src/export.ts
abhinav-phi: Addressed in 2abae42 with explicit limits checked before any content is retained: at most 1000 files, 1 MiB per file (via `stat`, never read first), and 8 MiB aggregate, each with a clear refusal error naming the limit.

Boundary coverage in `test/export.test.ts`: count 1000 passes while 1001 refuses (and writes nothing), per-file exactly 1 MiB passes while 1 MiB + 1 byte refuses, and aggregate exactly 8 MiB passes while 9 MiB refuses. Refusals are verified to leave no output file behind. I did not go the streaming route since the refusal path keeps the implementation (single joined document for both stdout and `--out`) intact; the limits sit far above any realistic hand-written scaffold while refusing far below heap-exhaustion territory.

abhinav-phi:
@theDakshJaitly polite ping for re-review on feat/export-command at 2abae423dd4cecca2c89ace54c5a297e15704656. P1/P2 are addressed in 2abae42. CI triage for run 35329123117: all green except release-performance finalize (job 105553777521, 15s); release-performance-attempt-1 (105549104146, 8m7s) and attempt-2 (105551416158, 8m24s) both passed. Finalize failed on confirmed budget_exceeded (e.g. maintenanceMs small wiki_refresh 581ms vs 186 budget, small wiki_rebuild 1192ms vs 181, large wiki_rebuild 1749ms vs 617). I tried gh run rerun 35329123117 --failed but got Must have admin rights, so could you rerun the failed finalize or advise? Thanks!

theDakshJaitly:
> @theDakshJaitly polite ping for re-review on feat/export-command at 2abae423dd4cecca2c89ace54c5a297e15704656. P1/P2 are addressed in 2abae42. CI triage for run 35329123117: all green except release-performance finalize (job 105553777521, 15s); release-performance-attempt-1 (105549104146, 8m7s) and attempt-2 (105551416158, 8m24s) both passed. Finalize failed on confirmed budget_exceeded (e.g. maintenanceMs small wiki_refresh 581ms vs 186 budget, small wiki_rebuild 1192ms vs 181, large wiki_rebuild 1749ms vs 617). I tried gh run rerun 35329123117 --failed but got Must have admin rights, so could you rerun the failed finalize or advise? Thanks!

Alright yes I'll look into it asap and also re-review

src/export.ts
```diff
@@ -0,0 +1,185 @@
...
+  out: string,
+  excludeReal?: string
+): void {
+  const target = resolve(config.projectRoot, out);
+  const targetReal = realpathBestEffort(target);
+  const protectedPaths = [...files, configPath];
+  for (const protectedPath of protectedPaths) {
+    const resolved = resolve(protectedPath);
+    if (excludeReal !== undefined && realpathBestEffort(resolved) === excludeReal) continue;
+    if (target === resolved || targetReal === realpathBestEffort(resolved)) {
```
theDakshJaitly, line 175: [P1] Reject existing non-export destinations before writing

The new guard fixes the reported Router/configuration/symlink examples, but its protected set contains only discovered Markdown and `config.json`. Other existing project state still falls through to `writeFileSync`: the built CLI accepted `--out .mex/events/decisions.jsonl`, exited 0, and replaced valid event history with the bundle. An existing README was also replaced. A hard link at `exports/scaffold.md` to the Router bypassed the realpath comparisons and overwrote the Router itself. Please refuse existing non-export destinations and preserve MEX-owned state regardless of the scanner's file list, with creation/replacement that cannot silently truncate an alias or a changed target. Keep any intended prior-export replacement exception narrowly scoped. Add preservation tests for event history and hard-linked source files.

src/export.ts
```diff
@@ -0,0 +1,185 @@
...
+        `Scaffold totals more than ${MAX_EXPORT_TOTAL_BYTES} bytes; ` +
+          `export supports at most ${MAX_EXPORT_TOTAL_BYTES} bytes in total.`
+      );
+    }
+  }
+
+  const bundle: string[] = ["# mex scaffold export", ""];
+  for (const file of files) {
+    const relativePath = toPosix(relative(config.scaffoldRoot, file));
+    bundle.push(`## ${relativePath}`, "", readFileSync(file, "utf-8").trimEnd(), "");
```
theDakshJaitly, line 141: [P2] Bound the actual file reads rather than only the earlier stat

The size preflight and this unrestricted read reopen each path separately. If a writer grows or replaces a file between them, the checked size no longer bounds the allocation. In a deterministic CLI probe, a preload hook grew a real source to 9 MiB immediately before this read, after the real stat calls returned; no stat result was falsified. Export still exited 0 and wrote a 9,437,267-byte bundle, exceeding both the 1 MiB per-file and 8 MiB aggregate limits. A FIFO at `.mex/context/live.md` also passes the size check with size 0 and hangs in the read; the probe required a timeout kill. Please reject non-regular inputs without blocking, read through a verified descriptor with actual byte ceilings and change checks, and accumulate actual bounded byte counts. Add coverage for source growth/replacement and non-regular inputs.

theDakshJaitly (changes requested):
Re-reviewed 2abae42 on current main. The original Router/configuration/symlink cases now preserve their bytes, repeated exports no longer include themselves, and the original oversized fixture is refused cleanly. The 70 existing focused tests, production build, workspace typecheck, and diff check pass. Two remaining gaps are documented inline: other existing project state can still be overwritten, and the byte limits do not bound the actual reads.

CI follow-up: I inspected the retained reports from [run 35329123117](https://github.com/mex-memory/mex/actions/runs/35329123117). Both green measurement jobs produced valid reports with `confirmation_required`; they did not pass the performance budgets. The final gate confirms five Wiki timing failures across distinct runners: small/medium Wiki refresh and rebuild, plus large Wiki rebuild. For example, small Wiki refresh measured 780.673 ms and 581.305 ms against a 186 ms budget. The finalizer only evaluates those saved reports, so rerunning that job alone would reuse the same failing evidence. After the code fixes, a meaningful retry needs fresh measurement jobs and a clean final gate on the final head. These reports alone do not establish that the export change caused the Wiki slowdown.

abhinav-phi:
Addressed both follow-ups in b3c6de5, verified with unit tests (16/17 locally - the 1 failure is the pre-existing Windows file-symlink privilege EPERM in the old symlink-alias test setup, unrelated) plus built-CLI probes:

P1 (existing non-export destinations): the guard no longer protects only scaffold+config. Any pre-existing path at --out (event history, README, hardlink/symlink/FIFO/dir) is refused before anything is read or written, unless it carries the bundle marker (previous export). New tests: decisions.jsonl-style state, README, hardlink-to-Router (same inode, invisible to realpath), existing directory. Built-CLI probes confirmed: decisions.jsonl, README, and hardlink-to-Router all refused with original bytes intact; normal and repeat exports still work.

P2 (bound actual reads): removed the stat-then-read split. Each file is opened O_NONBLOCK and read through its own descriptor with the per-file cap enforced on bytes actually transferred (fstat must report a regular file, so FIFOs are refused instead of hanging; the open pins the inode, so post-discovery growth/replacement cannot escape the cap). Aggregate cap is enforced on actual bytes before the document is allocated.

Writes are atomic: new files use exclusive wx create (EEXIST becomes the same refusal), previous-bundle overwrites re-verify the marker on the write descriptor before truncating.

CI note: the failing release-performance Finalize only re-evaluates the saved Wiki-timing reports from run 35329123117 (small/medium refresh+rebuild, large rebuild over budget) - rerunning it alone reuses that evidence. After this fix lands, the measurement jobs need a fresh run plus a clean final gate on the final head; those Wiki numbers do not implicate the export change by themselves.

@theDakshJaitly please re-review when you get a chance. Thanks!

src/export.ts
abhinav-phi: Done in b3c6de5 - any pre-existing path at --out that is not a previous export bundle (event history, README, hardlink/symlink/FIFO/dir) is now refused before anything is read or written. Verified with new regression tests (decisions-style state, README, hardlink-to-Router, directory) plus built-CLI probes with bytes intact. Please re-check when convenient.

src/export.ts
abhinav-phi: Done in b3c6de5 - the stat-then-read split is gone. Each file is opened O_NONBLOCK and read through its own descriptor with the per-file cap enforced on bytes actually transferred (fstat must report a regular file, so FIFOs are refused instead of hanging; the open pins the inode). Aggregate cap is enforced on actual bytes. New + existing bound tests pass (16/17 locally, the 1 failure being the pre-existing Windows symlink-privilege EPERM in old test setup). Please re-check when convenient.

abhinav-phi:
CI update on 1632584 (run 35453570839): check (22+24), hub-browser, storage-portability (macos+windows), and both perf measurement attempts are all green - including the fixed architecture-inventory test. The only red is the release-performance Finalize gate, and only on Wiki/graph/browser budgets: medium wiki_rebuild 359ms vs 301, small wiki_refresh 288ms vs 186, small wiki_rebuild 434ms vs 181, graph_refresh RSS 481MB vs 470MB, browserHeap home 6.76MB vs 4.93MB (fresh measurements, both attempts). This change touches only src/export.ts + export/architecture tests, which have no path to the wiki-rebuild/graph/browser hot paths, so I do not see a mechanism by which it could move those numbers - and main-branch CI is green today. @theDakshJaitly - does this perf gate block merge of the export change, or can it be waived/tracked separately as a budgets-vs-runners issue? Happy to do anything that would actually implicate this diff.

src/export.ts
```diff
@@ -0,0 +1,309 @@
...
+    if ((error as NodeJS.ErrnoException)?.code === "EEXIST") {
+      throw new Error(
+        `Refusing to export: "${out}" already exists and is not a previous export bundle. ` +
+          `Choose a different --out path.`
+      );
+    }
+    throw error;
+  }
+  try {
+    writeSync(fd, document, 0, "utf-8");
```
theDakshJaitly: [P2] Complete partial writes before reporting export success

This single `writeSync` call ignores the number of bytes actually written; `overwritePreviousBundle` has the same issue at line 149. A successful call can transfer only a prefix. With the real built CLI on macOS and a child-only `RLIMIT_FSIZE=4096`, a 20,083-byte export wrote just 4,096 bytes, exited 0, and printed `Wrote 2 scaffold file(s) ...`. Both new-file creation and previous-bundle replacement reproduced this without mocking writes. Use a helper that completes descriptor writes, or loop over the remaining UTF-8 bytes and propagate any failure. Only report success after the entire document is written, and add partial-write coverage for both branches.

src/export.ts
```diff
@@ -0,0 +1,309 @@
...
+  return join(realpathSync(current), ...missing);
+}
+
+/** Whether an existing path is a previous export bundle (marker prefix, bounded read). */
+function isPreviousExportBundle(target: string): boolean {
+  let fd: number | undefined;
+  try {
+    fd = openSync(target, "r");
+    const prefix = Buffer.alloc(BUNDLE_MARKER.length);
+    const read = readSync(fd, prefix, 0, prefix.length, 0);
```
theDakshJaitly: [P2] Refuse non-regular destinations without a blocking marker read

The new nonblocking/type checks protect scaffold inputs, but this destination probe still opens with blocking `"r"` before the existence refusal. With a real FIFO at `scaffold.md`, `mex export --out scaffold.md` waits for a writer in this call instead of refusing the destination; the built-CLI probe had to be killed after three seconds. Open the marker probe nonblocking and verify that its descriptor is a regular file before reading, and make the later previous-bundle descriptor check safe if the destination is replaced by a non-regular object. Add a bounded CLI test for a FIFO output path.

theDakshJaitly (changes requested):
The earlier overwrite and input-read findings are now resolved in the regression probes. The 73 existing focused tests, production build, workspace typecheck, and diff check pass. Please fix the two remaining output-handling P2 issues below before approval: incomplete writes currently report success, and a FIFO destination can hang during the previous-bundle check.

On the performance question: the final report for [run 35453570839](https://github.com/mex-memory/mex/actions/runs/35453570839) has one blocking violation, medium Wiki rebuild. The independent attempts measured 1393.517 ms and 359.254 ms against the 301 ms budget. The other raw Wiki/graph/browser crossings mentioned in the update are not additional final blocking violations. The green measurement jobs mean that reports were produced; the final gate is still failed. This evidence alone does not establish that export caused the slowdown, and no waiver is being made in this review. Rerunning only the finalizer would reuse the saved measurements; investigating runner variability needs fresh measurements and a comparable main baseline.

abhinav-phi:
@theDakshJaitly Both P2s fixed in 6848ce0: (1) new writeAllSync helper loops descriptor writes over remaining bytes and throws unless the full document lands — wired into both the new-file and overwrite branches, with failure coverage for both plus a short-write-completes test; (2) the marker probe and the overwrite descriptor now open O_NONBLOCK with a regular-file check, so a FIFO --out is refused instead of hanging — plus a bounded FIFO refusal test (skipped on Windows where FIFOs don't exist). Local: typecheck clean, export suite 20/21 (the one failure is the pre-existing symlink EPERM on Windows runners, unrelated). On the perf gate: the Wiki-rebuild crossing looks unrelated to this diff (export path only, no wiki/graph code touched) — happy to trigger whatever confirmation run you prefer, or leave the waiver call to you. PTAL, thanks!

theDakshJaitly:
Thanks, I'll review asap.

theDakshJaitly (approved):
All earlier findings are resolved: existing project state is preserved, actual scaffold reads are bounded, partial writes cannot report success, and FIFO inputs/destinations are refused without blocking. The FIFO regression test now creates a real fixture and uses a bounded CLI subprocess.

Validation: 72 focused tests, all workspace typechecks, the production build, and 12 independent CLI probes passed. A forced blocking-open regression is caught by the test timeout. Full CI is green on a0c1a92, including Node 22/24, browser, Windows/macOS, packaging, and release performance: https://github.com/mex-memory/mex/actions/runs/36024810155

PR #186, opened 2026-09-09
feat(drift): filter non-package dependency claims

### LINKED ISSUE #4: Smarter filtering for non-package dependency claims
## Problem

The claim extractor (`src/drift/claims.ts`) treats every `**bold**` word in stack/tech sections as a dependency claim. This produces false positives for things that are clearly not package names: multi-word phrases ("REST API"), all-caps acronyms ("AWS"), descriptive labels ("Frontend", "Backend", "Database Layer").

## Where to look

- `src/drift/claims.ts` — the `strong` node visitor (~line 105)
- `src/drift/checkers/dependency.ts` — where claims are checked

## What to do

Add heuristics to filter out claims that aren't real package names before they reach the dependency checker. Ideas:

- Skip multi-word phrases that don't look like scoped packages
- Skip all-caps words (likely acronyms, not packages)
- Skip common architectural/descriptive terms

This is complementary to expanding `KNOWN_RUNTIMES` — that's a blocklist approach, this is pattern-based filtering.

## Acceptance criteria

- Fewer false positive `DEPENDENCY_MISSING` warnings on real-world READMEs
- Real package names still get checked correctly
- Add test cases

abhinav-phi:
Resolves #4.

## Problem

The claim extractor treated every bold word in a stack/tech section as a dependency claim, so **AWS**, **REST API**, and **Database Layer** were checked against `package.json` and reported `DEPENDENCY_MISSING` — fewer false positives wanted, per the issue.

## What

Three filters, layered after the existing `PACKAGE_NAME` shape check:

1. **All-caps words** are acronyms (`AWS`, `REST`, `JWT`) — no package is styled that way.
2. **Multi-word phrases** are descriptive labels (`REST API`, `Database Layer`); scoped names (`@scope/pkg`) keep an explicit exemption since npm scopes contain a slash but never spaces.
3. **A curated architectural-label blocklist** (`frontend`, `backend`, `middleware`, `auth`, `gateway`, ...) as the complement the issue describes — the patterns cannot catch a single-word label like "Frontend".

Design note per the issue's framing: this stays pattern-based (zero-config), not structured output; the maintainer's earlier comment on this issue settled that direction.

## Tests

Four new cases in `test/claims.test.ts`: acronyms dropped, multi-word phrases dropped, architectural labels dropped while real packages (`Express`, `@scope/pkg`) and mixed-case names (`pino-http`) survive. All 30 claims tests pass; `npm run typecheck` green.

### DIFF
--- a/src/drift/claims.ts
+++ b/src/drift/claims.ts
@@ -38,6 +38,24 @@ const DOTTED_KEY_WITH_SLASH = /^[A-Za-z0-9_-]+(?:\.[A-Za-z0-9_-]+)+\/[A-Za-z0-9_
  */
 const PACKAGE_NAME = /^@?[A-Za-z0-9][A-Za-z0-9._/-]*$/;
 
+/**
+ * Architectural/descriptive terms that pass PACKAGE_NAME but are not
+ * installable packages — the blocklist complement to the structural
+ * heuristics above. One word per label; real packages never collide
+ * (npm names "api" or "database" are far rarer than the false positives
+ * these labels cause in stack docs).
+ */
+const ARCHITECTURAL_LABELS = new Set([
+  "frontend", "backend", "fullstack", "full-stack",
+  "database", "database layer", "storage layer", "data layer",
+  "api", "api layer", "service layer",
+  "middleware", "infrastructure", "infra", "platform",
+  "auth", "authentication", "authorization",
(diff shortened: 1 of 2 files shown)


abhinav-phi:
hi this filters non package dependency claims from drift results

framework level mentions no longer show up as false dependency claims

let me know if the filter needs tightening

Issue #143, opened 2026-08-26
MISSING_PATH false positives: negation only heading-scoped, non-path tokens treated as paths, dot-dir glob blind spot


m3:
## Summary

`mex check` (0.7.0-0.7.2) emits MISSING_PATH **errors** for files that are intentionally absent or were never paths at all, which zeroes drift scores on healthy repos.

## Three defects

1. **Negation is heading-scoped only** (`isNegatedSection`). Bullets describing *deletions* inside normal sections — e.g. "Deleted orphaned files: `check_commands.js`, ..." — still produce path claims for intentionally-absent files.
2. **Non-path tokens extracted as paths**: ELO-style deltas (`+/-N`), `overall/overall`, middot lists (`8ball/pro · 9ball/pro`), globs (`web/routes/admin/*.js`), trailing dirs (`screenshots/`).
3. **Dot-directory blind spot**: the fallback filename glob lacks `dot: true`, so a backticked `deploy.yml` living in `.github/workflows/` never matches, producing a false error.

## Repro and evidence

Full root-cause plus a verified fix against our installed 0.7.0:

- added inline line-level negation context (deleted / removed / dropped / retired / orphaned / unreferenced / no longer)
- NON_PATH_TOKEN filter for deltas, middot lists, globs and trailing dirs before claim extraction
- negated/implausible claims demote MISSING_PATH from error to warning (visible as noise, not score killers)
- `dot: true` on the fallback filename glob

Verified across a 10-project estate: one project went from 23 MISSING_PATH errors (score 0) to 0 errors; 7 known-good projects stayed 100/100.

Happy to send the patch diff or open a PR if useful.

theyashasvipandey:
Thanks — good report, all three confirmed against `main` (0.7.3). 
PR very welcome.

One thing to settle first: the error→warning demotion. The other three changes fix
detection and I'd take them as-is, but demotion doesn't remove a false positive, it
just makes it cheaper — costs are flat per issue with no cap, so it takes ~34 to floor
the score instead of ~10. Bigger bucket, same leak.

My instinct: if `NON_PATH_TOKEN` is confident a token isn't a path, drop it before it
becomes a claim — no issue, no points. Anything that survives into a claim deserves
`error`. Is there a case where the filter is confident enough to demote but not
confident enough to drop?

PR #210, opened 2026-09-19
fix(drift): do not treat version-shaped refs as missing paths

### LINKED ISSUE #201: MISSING_PATH: a version-shaped reference like release/2.1.0 is reported as a missing file
`mex check` reports a MISSING_PATH error for references shaped
`<word>/<version>` — git branch names and pinned runtimes, not file paths:

```
✗ MISSING_PATH:123 Referenced path does not exist: release/2.1.0
```

`src/drift/checkers/path.ts` has a guard, `isUnrootedReference`, whose job is to
treat this kind of token as prose: if nothing by the name of the first segment
exists, it is not a real relative path. The guard never runs here, because it
returns early on anything that looks like it has a file extension:

```js
if (/\.[A-Za-z0-9]+$/.test(trimmed)) return false;
```

A trailing version number matches that pattern — `.1` in `2.1.0` reads as an
extension — so the value skips the prose check and is reported missing.

The inconsistency is visible within a single document. These differ only by a
trailing word:

```
release/2.1.0      FLAGGED        release/2.1.0-rc   ignored as prose
api/v1.2           FLAGGED        api/v1             ignored as prose
python/3.11        FLAGGED
node/20.11         FLAGGED
```

`release/2.1.0` is the git-flow standard branch name, so this affects any
project that documents a release branch, and `python/3.11` or `node/20.11` is
the ordinary way to write a pinned runtime in a stack document.

**To Reproduce**

`main` already carries a failing case for this, skipped:

```
src/drift/__tests__/path-false-positives.test.ts
  it.skip("does not claim a version-shaped branch name", …)
```

Remove the `.skip` to reproduce.

**Expected behavior**

A trailing numeric segment is a version, not a file extension, so the value
should reach the existing prose guard and be dropped rather than reported.

**One boundary worth knowing**

Making the extension check require an alphabetic character is only half of it.
Once the value reaches `isUnrootedReference`, that guard clears it *only if the
first path segment does not exist on disk*. So `api/v1.2` in a repository that
really has an `api/` directory will still be reported. The skipped test covers
the common case; the narrower one can be left alone or handled separately.

dyk1454683243-sudo:
## What

Tighten the `isUnrootedReference` extension guard so a trailing **numeric-only** segment is not treated as a file extension.

`release/2.1.0` used to match `/\.[A-Za-z0-9]+$/` on `.0`, skip the prose/unrooted check, and report `MISSING_PATH`. The guard now requires an alphabetic character in that suffix, so git-flow / pinned-runtime tokens (`release/2.1.0`, `python/3.11`, `node/20.11`) reach the existing first-segment check.

Unskips `does not claim a version-shaped branch name` and adds a small regression that `docs/foo.bar` still stays a real path claim.

The narrower `api/v1.2` case (first segment exists on disk) is unchanged.

## Why

Fixes #201

`mex check` was flagging branch names and pinned runtimes as missing files whenever the last dotted segment was digits-only.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/Tooling

## How to test

1. `npx vitest run src/drift/__tests__/path-false-positives.test.ts`
2. Confirm the previously skipped `does not claim a version-shaped branch name` case is active and green.
3. Existing #143 path-false-positive cases should stay green.
4. `npm run typecheck`

## Checklist

- [x] Focused drift path tests pass (`npx vitest run src/drift` — 12/12)
- [ ] Full `npm test` — not used as a gate here; this environment’s Node 22.14.0 SQLite is built without FTS5, which fails graph/wiki suites unrelated to this change

## Code-graph changes

- [ ] This PR targets `main`


src/drift/checkers/path.ts
```diff
@@ -94,7 +94,10 @@ function isUnrootedReference(
   const trimmed = value.replace(/\/+$/, "");
   const isDirectoryRef = trimmed !== value;
   if (!trimmed.includes("/") && !isDirectoryRef) return false;
-  if (/\.[A-Za-z0-9]+$/.test(trimmed)) return false;
+  // A trailing numeric-only segment is a version (`release/2.1.0`,
+  // `python/3.11`), not a file type. Require an alphabetic character so
+  // `.ts` still counts as an extension and those tokens reach the prose check.
+  if (/\.[A-Za-z0-9]*[A-Za-z][A-Za-z0-9]*$/.test(trimmed)) return false;
```
theDakshJaitly, line 100: [P2] Preserve missing files with numeric extensions

A numeric extension does not necessarily make the complete filename a version. For example, when `manuals/` exists at neither the project nor scaffold root, a documented `manuals/mex.1` now reaches the unrooted-reference check and is silently discarded. The base commit correctly reports `MISSING_PATH`. The same regression occurs for `manuals/admin.8` and `logs/server.log.1`.

Reproduction to add to `path-false-positives.test.ts`, using its existing helper:

```ts
expect(missingPaths("# Docs\n\n- See `manuals/mex.1`\n"))
  .toEqual(["manuals/mex.1"]); // actual on this PR: []
```

Please recognize a complete version-shaped final component before applying the exemption, and retain ordinary numeric-extension filenames. Keep the existing-parent and explicitly relative path behavior intact.

theDakshJaitly (changes requested):
This fixes the version-reference examples from #201, but the numeric-suffix exemption also suppresses real missing-file reports.

Please narrow the exception to complete version-shaped final components, such as `2.1.0` or `v1.2`, instead of treating every numeric extension as a version. Preserve `MISSING_PATH` for ordinary filenames such as `manuals/mex.1`, `manuals/admin.8`, and `logs/server.log.1` when their first directory is absent. Add regression tests for these cases alongside the version-reference tests; the inline comment includes a reproduction using the existing test helper.

Validation on `bced9b5`: 148 focused drift/checker/claim tests passed, workspace typecheck passed, and `git diff --check` passed. Additional tests using Markdown extraction and `checkPaths` reproduced all three missing-file regressions; the same assertions pass on the base commit. Full repository and performance suites were not run.

dyk1454683243-sudo:

Issue #205, opened 2026-09-19
An interrupted graph rebuild leaves a full-size candidate file that nothing ever removes


theyashasvipandey:
`mex graph rebuild` builds into `graph.db.candidate-<token>` and publishes it by
atomic rename. If the process dies partway — Ctrl+C, a CI timeout, a crash — the
candidate stays on disk permanently. No later command removes it, and no command
mentions it.

**To reproduce**

Kill a running `mex graph rebuild`, then run each maintenance command in turn:

```
$ ls .mex/
graph.db                                             6,774,784
graph.db.candidate-ea0c83460fd19a3425f426cd7a32ec…   6,774,784
graph.db.lock                                              112
```

| after | candidate present? | notes |
| --- | --- | --- |
| `mex graph status` | yes | reports `fresh`, says nothing about it |
| `mex graph repair` | yes | |
| `mex graph refresh` (successful) | yes | |
| `mex graph rebuild` (full, successful) | **yes** | the stale `.lock` *is* cleared; only the candidate leaks |

The candidate is a byte-for-byte duplicate of the store, so every interruption
costs one full store on disk. It is invisible unless someone lists `.mex/`.

**Cause**

`src/graph/maintenance.ts` calls `cleanupOwnedDatabase(paths, candidatePath)`
only from within the run that created the candidate — its catch/finally paths.
A killed process never reaches that code, and there is no sweep: the file
contains no `readdirSync` or `globSync` of the graph directory at all, so no
later run can discover a previous run's leftovers.

The same applies to the other two owned prefixes in `OWNED_DATABASE_PREFIXES` —
`graph.db.rollback-` and `graph.db.recovery-`. Planting one of each confirms
both survive status, repair, refresh and a successful rebuild.

**Impact**

One full copy of the store leaked silently per interruption. On the 240 MB store
in #140 that is 240 MB each time, and that report describes builds being stopped
by a 30-minute timeout — exactly this path.

**Note for anyone triaging this alongside #140**

On 0.8.x an interrupted rebuild does **not** strand a WAL; it strands a
candidate plus a lock. The two failure modes look alike and are not:

| | stranded `graph.db-wal` | orphan `graph.db.candidate-*` |
| --- | --- | --- |
| what's wrong | status misreports a healthy store | a full-size copy stays on disk |
| cost | confusion | disk |
| fix today | `mex graph repair` | delete it by hand |
| needs | wording | cleanup + a warning |

The wording half is filed separately as #204.

**Suggested fix**

[...]

PR #213, opened 2026-09-19
fix(graph): do not print uninspected status as measured zeros

### LINKED ISSUE #204: graph status prints uninspected fields as measured zeros — a stranded graph.db-wal reads as an empty graph
When `mex graph status` finds an active sidecar next to the store it skips
immutable inspection — correctly — but still prints placeholder values in the
same shape it prints measurements. Three of the five lines are then false.

**To reproduce**

On a 111-file TypeScript repository with a healthy, freshly indexed store:

```
$ mex graph status
Graph status: stale
Repository: main @ df89810df4f5
Last successful index: 2026-09-13T04:33:14.973Z
Sources: 1 changed (0 added, 0 modified, 1 deleted)
Parse health: 111 ok, 0 partial, 0 failed
```

Strand a non-empty WAL beside the store — an interrupted write leaves exactly
this — and run the same command against the same untouched database:

```
$ mex graph status
Graph status: degraded
Repository: main @ df89810df4f5
Last successful index: never
Sources: 0 changed (0 added, 0 modified, 0 deleted)
Parse health: 0 ok, 0 partial, 0 failed
WARNING GRAPH_INDEX_SIDECAR_ACTIVE: Graph maintenance or recovery is active (graph.db-wal); immutable inspection was skipped.
Next: mex graph repair
```

Nothing about the store changed between those two runs. `mex graph repair`
restores `111 ok` and the real index timestamp immediately.

* `Parse health: 0 ok` reads as "nothing parsed". It means "not inspected".
* `Last successful index: never` is false — the store was indexed on 09-13.
* `Sources: 0 changed` reads as "up to date". Nothing was compared.

**`--json` is partly affected.** `lastSuccessfulIndexAt`, `indexedAt`,
`schemaVersion` and the version fields all come back `null`, which is honest.
But `parseHealth` and `changes` come back as fully-populated zero objects:

```json
"parseHealth": { "total": 0, "ok": 0, "partial": 0, "failed": 0,
                 "failedPaths": [], "failedPathsTruncated": false },
"changes":     { "total": 0, "added": [], "modified": [], "deleted": [], ... }
```

A consumer cannot distinguish that from a store that genuinely parsed nothing.

**Cause**

* `printStatus` (`src/graph/cli-graph.ts:181`) prints all five lines
  unconditionally, with `?? "never"` supplying the index-time placeholder.
* `src/graph/status.ts` has **16 early-return paths** that fill parse health
  with `emptyParseHealth()`. The WAL is only the case reproduced here — an
  unreadable sidecar, a containment failure or a failed invariant audit all
  print the same zeros.

**Expected behavior**

Distinguish "not inspected" from "measured zero". In text, something like
`Parse health: not inspected (graph.db-wal present)`, and suppress or mark the
index-time and sources lines on the same paths — that half is contained to
`printStatus`.

For JSON, prefer an **additive** `inspected: false` (or equivalent) over making
`parseHealth` and `changes` nullable. `parseHealth` alone has ~163 non-test
references, including invariant checks in `src/graph/read-session.ts` that
compare its counts against the file list, so widening the type is a much larger
change than the reporting bug warrants. An additive flag breaks no consumer.

**Additional context**

This was diagnosed publicly in #140 — "that placeholder reads like 'zero files
parsed' when it actually means 'not inspected'; we'll fix the wording" — but
nothing on `main` changes it and no issue tracks it. Filing so it is not lost.
The severity is that a clone's `0 ok` was read there as a failed graph when the
store was intact.

dyk1454683243-sudo:
## What

When immutable graph inspection is skipped (a stranded `graph.db-wal` is the filed case), `mex graph status` no longer prints placeholder zeros that look like real measurements.

`GraphStatus` now carries an additive `inspected` flag. Early-return paths that skip opening or reading the store set `inspected: false`. `printStatus` then labels parse health, last successful index, and sources as not inspected instead of `0 ok` / `never` / `0 changed`. `--json` keeps the existing zero `parseHealth` and `changes` objects; consumers distinguish “not inspected” from an empty graph with the new flag.

## Why

Fixes #204

A healthy store with a non-empty WAL sidecar was reading as `Parse health: 0 ok` and `Last successful index: never` even though the database was unchanged. Those zeros are placeholders from `emptyParseHealth()`, not measurements.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/Tooling

## How to test

1. `npx vitest run src/graph/__tests__/cli-graph.test.ts src/graph/__tests__/status.test.ts`
2. Confirm sidecar-active status has `inspected === false` and human output does not claim `0 ok`.
3. Confirm a healthy inspected graph still prints real parse-health counts and `inspected === true`.
4. `npm run typecheck`
5. `npm test`

## Checklist

- [x] Focused graph tests pass (`npx vitest run src/graph/__tests__/cli-graph.test.ts src/graph/__tests__/status.test.ts` — 58 passed, 1 skipped)
- [x] Tests pass (`npm test`) — 4168 passed, 4 skipped; 7 failures were timeouts / setup-signing flakes unrelated to this status-reporting change and reproduced or vanished on isolated reruns

`parseHealth` and `changes` stay required non-null objects. `inspected` is optional and defaults to true on fully inspected paths, so existing consumers remain valid.

## Code-graph changes

- [x] This PR targets `main`
- [x] No graph identity, reconciliation, schema, or drift-semantics changes are included, or a `core / discuss-first` issue is linked above

This is a status-reporting fix only. It does not change indexing, publication, or #205 orphan-candidate work.


src/graph/status.ts
```diff
@@ -1166,9 +1167,16 @@ function graphStatus(input: {
     parseHealth: input.parseHealth,
     changes: input.changes,
     diagnostics: input.diagnostics,
+    inspected: input.inspected ?? true,
```
theDakshJaitly, line 1170: [P2] Do not mark the snapshot-mismatch placeholders as inspected

The `GRAPH_SNAPSHOT_CONTENT_MISMATCH` return above still calls `graphStatus(...)` without overriding this default. Although it has measured `parseHealth`, it supplies `changesWithoutIndex(...)` (all-zero placeholders) and omits index timestamps, so `inspected: true` is incorrect for the aggregate result.

Reproduced on this head:

1. Build a graph for one TypeScript file and retain the successful-index timestamp.
2. Change the snapshot's `sourceCorpusDigest` to `"0".repeat(64)` while preserving the rest of the snapshot.
3. Modify the live source file.
4. Run `runGraphStatus` in text and JSON modes.

Text still prints `Last successful index: never` and `Sources: 0 changed (0 added, 0 modified, 0 deleted)`. JSON contains `inspected: true`, `lastSuccessfulIndexAt: null`, and `changes.total: 0`, despite the known prior index and source edit. The existing corruption diagnostic is correct, but the timestamp and source comparison were not measured.

Please mark this return uninspected (or distinguish the unavailable fields explicitly) and add the corresponding regression test. This is a missed case in the reporting fix, not a request to relax the corruption check.

theDakshJaitly (changes requested):
This fixes the stranded-WAL case, but the `GRAPH_SNAPSHOT_CONTENT_MISMATCH` early return still reports unmeasured fields as facts. It has read parse-health counts, but has not compared sources and does not populate the index timestamps. The new default therefore labels its placeholder `changes` and timestamps as `inspected: true`.

Please handle this partial-inspection return too: mark the unavailable fields as not inspected, using the existing flag conservatively or an equivalent explicit distinction. Add a regression that builds a graph, changes its recorded snapshot digest, modifies a source file, and verifies both text and JSON output.

Validation on `254127f`: 200 existing focused tests passed across graph status, CLI, graph integration/freshness/degradation, Hub services/contracts, and reporter/TUI. Workspace typecheck and `git diff --check` passed. Additional controls confirm fresh/missing-index behavior and stranded-WAL text/JSON output, including unchanged database/WAL bytes. The snapshot-mismatch regression fails as described in the inline comment. Full repository, packaging, and performance suites were not rerun.

theDakshJaitly:
Closing as superseded by #215, which has now merged. Both PRs address #204 by distinguishing uninspected graph-status placeholders from measured values. The merged implementation also covers the snapshot-content-mismatch case raised in review and the unavailable-status fallback, with regression coverage and passing CI. Thanks for contributing this fix.

PR #214, opened 2026-09-19
fix(graph): sweep orphan owned-prefix files after leased maintenance

### LINKED ISSUE #205: An interrupted graph rebuild leaves a full-size candidate file that nothing ever removes

dyk1454683243-sudo:
## What

Interrupted `mex graph rebuild` can leave a full-size `graph.db.candidate-<token>` (and the same for `rollback-` / `recovery-`). `cleanupOwnedDatabase` only removes the path the current run created, so a killed process leaked a store-sized file forever.

After `acquireGraphMaintenanceLease` succeeds, refresh / repair / rebuild now readdir `.mex/` and remove regular files whose basename starts with an `OWNED_DATABASE_PREFIXES` entry, except the live in-progress candidate for that run (if any). The sweep is exclusive-owner only — it does not run before the lease.

`mex graph status` is still read-only: it emits `GRAPH_INDEX_ORPHAN_OWNED_DATABASE` while one of those files is present and does not delete anything.

This is independent of stranded-WAL status wording (#204).

## Why

Fixes #205

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/Tooling

## How to test

1. Plant `graph.db.candidate-*`, `graph.db.rollback-*`, and `graph.db.recovery-*` under `.mex/` plus an unrelated file.
2. Run `mex graph status` — it stays `fresh` (when the store is) and prints a warning; the files remain.
3. Run `mex graph refresh` (or `repair` / `rebuild`) — the owned-prefix leftovers are gone; the unrelated file remains.
4. Hold a lease and retry refresh — the command fails `GRAPH_MAINTENANCE_LOCKED` and the leftovers stay.

```
npx vitest run src/graph/__tests__/maintenance.test.ts src/graph/__tests__/status.test.ts
npm run typecheck
```

## Checklist

- [x] Focused graph maintenance + status tests pass (`npx vitest run src/graph/__tests__/maintenance.test.ts src/graph/__tests__/status.test.ts` — 94 passed, 1 skipped)
- [ ] Tests pass (`npm test`) — full suite not used as a gate here

## Code-graph changes

- [x] This PR targets `main`
- [x] A linked issue agrees on the bounded extractor/resolver scope
- [x] No graph identity, reconciliation, schema, or drift-semantics changes are included, or a `core / discuss-first` issue is linked above


dyk1454683243-sudo:

Issue #202, opened 2026-09-19
MISSING_PATH: six notations and exemptions that are not drift


theyashasvipandey:
After #143, `mex check` still reports MISSING_PATH errors for things that are not
drift. Measured across five scaffolds with real `.mex/` directories, one of them
loses ~90 points to notations that were never paths.

Grouped because they share `src/drift/claims.ts` and
`src/drift/checkers/path.ts`. Happy to split further. A seventh defect in the
same area — version-shaped references such as `release/2.1.0` — is filed
separately as a starter task in #201.

### A. Tokens that were never file paths

| # | Flagged | Document actually says | Cause |
|---|---|---|---|
| 1 | `src/auth/login.validateToken` ×3 | "the `qualified_name` string (**e.g.** …)" | symbol notation read as a path |
| 2 | `.d.ts` ×2 | "the SDK's CJS `.d.ts`" | `EXTENSION_ONLY` is `/^\.[A-Za-z0-9]+$/` — matches `.ts`, misses the compound form |
| 3 | `./x`, `x.ts`, `x/index.ts` | "(`./x` may be `x.ts` or `x/index.ts`); **at most one really exists**" | hypothetical examples |
| 4 | `tests/grounding\|traversal\|edges\|store.test.ts` | "The existing unit tests in …" | alternation shorthand |

### B. Things that exist but are reported missing

**5. Scoped packages with a strict `exports` map.** `pathExists` probes
`${pkgName}/package.json`, which modern packages do not expose:

```
@scope/pkg/package.json  ->  FAILS: ERR_PACKAGE_PATH_NOT_EXPORTED
@scope/pkg               ->  RESOLVES
```

Observed on a package that is installed, in `dependencies`, and importable.
Resolving the bare specifier would fix it.

**6. The gitignore exemption is inert for directory-only rules.** Trailing-slash
patterns match directories only, and git cannot classify a path that does not
exist. mex writes exactly such a rule into every scaffold —
`src/setup/ignore.ts` has `["graph.db*", "wiki.db*", "local/"]`. In a clean repo
with those rules:

```
# .mex/local absent — the only case MISSING_PATH can fire
$ printf '.mex/local\n.mex/graph.db\n' | git check-ignore --stdin
.mex/graph.db                      <- .mex/local NOT exempt

# .mex/local present — already skipped by existsSync
.mex/local
.mex/graph.db
```

So the exemption fires only when it is not needed. `graph.db*` escapes because it
[...]

PR #217, opened 2026-09-19
fix(drift): resolve scoped packages by bare specifier

dyk1454683243-sudo:
## What

Change scoped-package existence checks in `pathExists` to resolve the bare specifier (`@scope/pkg`) instead of `${pkgName}/package.json`.

Modern packages with a strict `exports` map often do not export `./package.json`. Node then throws `ERR_PACKAGE_PATH_NOT_EXPORTED` even when the package is installed, listed in dependencies, and importable via the bare name. The workspace-name fallback is unchanged.

## Why

**Addresses #202 (bullet 5 only).**

Do **not** treat this as `Fixes #202` or `Closes #202` — that would auto-close the whole mega-issue.

Remaining #202 bullets stay open for follow-ups:

1. Symbol notation (`src/auth/login.validateToken`)
2. Compound extensions (`.d.ts`) — owned by #216
3. Hypothetical examples (`./x`, `x.ts`, `x/index.ts`)
4. Alternation shorthand (`tests/grounding\\|traversal\\|…`)
6. gitignore trailing-slash directory exemption

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/Tooling

## How to test

1. `npx vitest run src/drift`
2. Confirm an installed scoped package that blocks `./package.json` export is not `MISSING_PATH`.
3. Confirm a missing scoped name still reports `MISSING_PATH`.
4. Confirm existing workspace-name fallback still works (`test/checkers.test.ts` workspace cases).
5. `npm run typecheck`

## Checklist

- [x] Focused drift tests pass (`npx vitest run src/drift` — 12 passed, 1 skipped `#201` version-shaped case left untouched)
- [x] Path-checker unit tests pass (`npx vitest run test/checkers.test.ts` — includes the blocked-`package.json` export case and a still-missing scoped name)
- [x] Typecheck passes (`npm run typecheck`)
- [ ] Tests pass (`npm test`) — full suite is not the gate for this one-line resolve change
- [x] No breaking changes (or documented below)
- [ ] Tested locally with a real project

## Code-graph changes

- [ ] This PR targets `main`
- [ ] A linked issue agrees on the bounded extractor/resolver scope
- [ ] The change follows the frozen `LanguageExtractor` or `FrameworkResolver` interface
- [ ] A focused fixture and assertions for the expected node/edge shape are included
- [ ] Any new grammar WASM, extension mapping, extractor, or resolver is registered
[...]

PR #216, opened 2026-09-19
fix(drift): treat compound extensions like .d.ts as not-paths

dyk1454683243-sudo:
## What

Widen `EXTENSION_ONLY` so compound file-extension tokens such as `` `.d.ts` ``, `` `.d.mts` ``, and `` `.d.cts` `` are treated as extension-only prose and never become path claims.

The previous pattern `/^\.[A-Za-z0-9]+$/` matched `.ts` / `.yaml` but not extra dotted segments. Inline `` `.d.ts` `` still hit `KNOWN_EXTENSIONS` (ends with `.ts`) and was not filtered by `isNotAPath`, so the SDK prose example in #202 became `MISSING_PATH` twice.

Real paths with extensions stay claims: `src/index.ts`, `.mex/ROUTER.md`, `.github/workflows/ci.yml`.

## Why

**Addresses #202 (bullet 2 only).**

Do **not** treat this as `Fixes #202` or `Closes #202` — that would auto-close the whole mega-issue.

Remaining #202 bullets stay open for follow-ups:

1. Symbol notation (`src/auth/login.validateToken`)
3. Hypothetical examples (`./x`, `x.ts`, `x/index.ts`)
4. Alternation shorthand (`tests/grounding\\|traversal\\|…`)
5. Scoped-package resolve (`@scope/pkg/package.json` vs bare specifier)
6. gitignore trailing-slash directory exemption

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/Tooling

## How to test

1. `npx vitest run src/drift`
2. Confirm `` `.d.ts` `` / `` `.d.mts` `` / `` `.d.cts` `` produce zero path claims and zero `MISSING_PATH`.
3. Confirm existing #143 cases stay green, and real extension paths (`src/index.ts`, `.mex/ROUTER.md`, `.github/workflows/ci.yml`) are still checked.
4. `npm run typecheck`

## Checklist

- [x] Focused drift tests pass (`npx vitest run src/drift` — 12 passed, 1 skipped `#201` version-shaped case left untouched)
- [x] Typecheck passes (`npm run typecheck`)
- [ ] Tests pass (`npm test`) — full suite is not the gate for this one-regex change
- [x] No breaking changes (or documented below)
- [ ] Tested locally with a real project

## Code-graph changes

- [ ] This PR targets `main`
- [ ] A linked issue agrees on the bounded extractor/resolver scope
[...]

PR #218, opened 2026-09-19
fix(drift): honor trailing-slash gitignore rules for absent paths

dyk1454683243-sudo:
## What

Honor directory-only gitignore rules when collecting ignored paths for `MISSING_PATH`.

`collectIgnoredPaths` asked Git via `git check-ignore --stdin`. Trailing-slash patterns such as `local/` match directories only, and Git cannot classify a path that does not exist. On a clean checkout `.mex/local` was therefore **not** ignored — yet that is the only case `MISSING_PATH` can fire (when the directory is present, `existsSync` already skips). `graph.db*` already escaped because it is a glob with no trailing slash.

The ignore-collection path now also asks Git about:

- the directory form (`.mex/local/`)
- ignored parent prefixes (so a file under an absent ignored directory stays exempt)

The original claim is treated as ignored when any of those queries match. Real missing non-ignored paths still error.

## Why

**Addresses #202 (bullet 6 only).**

Do **not** treat this as `Fixes #202` or `Closes #202` — that would auto-close the whole mega-issue.

Remaining #202 bullets stay open for follow-ups:

1. Symbol notation (`src/auth/login.validateToken`)
2. Compound extensions (`.d.ts`) — owned by #216
3. Hypothetical examples (`./x`, `x.ts`, `x/index.ts`)
4. Alternation shorthand (`tests/grounding|traversal|...`)
5. Scoped-package resolve (`@scope/pkg/package.json` vs bare specifier) — owned by #217

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/Tooling

## How to test

1. `npx vitest run src/drift`
2. `npx vitest run test/checkers.test.ts`
3. Confirm an absent `.mex/local` (and a path under it) is not `MISSING_PATH` when `local/` is in `.mex/.gitignore`.
4. Confirm a present ignored path is still skipped, a missing non-ignored path still errors, and `.mex/graph.db` / `graph.db*` stay exempt.
5. Confirm the existing no-ignore `.mex/local/` case still reports `MISSING_PATH`.
6. `npm run typecheck`

## Checklist

- [x] Focused drift tests pass (`npx vitest run src/drift` — 15 passed, 1 skipped `#201` version-shaped case left untouched)
- [x] Path-checker unit tests pass (`npx vitest run test/checkers.test.ts` — 111 passed, including absent `local/` exemption, present ignored path, non-ignored missing path, and `graph.db*`)
- [ ] Tests pass (`npm test`) — full suite is not the gate for this ignore-collection change

## Code-graph changes

- [ ] This PR targets `main`

This is a drift path-ignore check only. It does not change graph identity, extractors, or the other #202 bullets.


dyk1454683243-sudo:

PR #219, opened 2026-09-19
fix(drift): ignore version-shaped path references

### LINKED ISSUE #201: MISSING_PATH: a version-shaped reference like release/2.1.0 is reported as a missing file

Voyagerroc-Lab:
## Summary
- Let numeric version suffixes such as `2.1.0` reach the existing unrooted-reference guard instead of treating `.0` as a filename extension.
- Enable the existing skipped regression case for `release/2.1.0` and `python/3.11`.

Closes #201

## Validation
- `npm test -- src/drift/__tests__/path-false-positives.test.ts` — 11 passed.
- `npm run typecheck` — passed after building the workspace contracts package.
- `git diff --check` — passed.
- A full `npm test` attempt encountered unrelated Windows symlink permission, file-mode, and graph-snapshot test failures; it was stopped after those failures. The focused path suite is clean.


src/drift/checkers/path.ts
```diff
@@ -94,7 +94,7 @@ function isUnrootedReference(
   const trimmed = value.replace(/\/+$/, "");
   const isDirectoryRef = trimmed !== value;
   if (!trimmed.includes("/") && !isDirectoryRef) return false;
-  if (/\.[A-Za-z0-9]+$/.test(trimmed)) return false;
+  if (/\.[A-Za-z][A-Za-z0-9]*$/.test(trimmed)) return false;
```
theDakshJaitly, line 97: [P2] Preserve real numeric and digit-leading file extensions

When the first directory exists at neither the project nor scaffold root, this guard lets two classes of real filenames reach the unrooted-reference check and disappear from the report: numeric extensions (`manuals/mex.1`, `manuals/admin.8`, `logs/server.log.1`) and extensions that begin with digits but contain letters (`backups/release.7z`, `models/widget.3mf`). All five produce `MISSING_PATH` on the base commit and no issue on this PR.

Reproductions to add to `path-false-positives.test.ts`, using its existing helper:

```ts
expect(missingPaths("# Docs\n\n- See `manuals/mex.1`\n"))
  .toEqual(["manuals/mex.1"]); // actual: []
expect(missingPaths("# Assets\n\n- See `backups/release.7z`\n"))
  .toEqual(["backups/release.7z"]); // actual: []
```

Please narrow the version exemption to complete version-shaped final components and preserve ordinary file extensions, including those starting with digits. The first-directory check should continue to apply to genuine version references.

theDakshJaitly (changes requested):
This fixes the version-reference examples from #201, but the new extension guard also hides real missing files. Two cases need to be preserved before merging:

1. Numeric extensions on ordinary filenames: `manuals/mex.1`, `manuals/admin.8`, and `logs/server.log.1` are now silently ignored when the first directory is absent. The exception should recognize a complete version-shaped final component, rather than treating every numeric extension as a version.
2. Extensions beginning with digits: `backups/release.7z` and `models/widget.3mf` are also silently ignored under the same conditions because the regex requires the first extension character to be a letter. These are file extensions and must retain their previous missing-path behavior.

Please add absent-parent regression tests for both classes, while keeping the version-reference tests and existing-parent/explicit-relative behavior intact. The inline comment includes reproductions using the existing test helper.

Validation on `10f51a4`: 147 focused drift/checker/claim tests passed, workspace typecheck passed, and `git diff --check` passed. Additional tests using Markdown extraction and `checkPaths` reproduced all five regressions; the same assertions pass on the base commit. Full repository and performance suites were not run.

PR #220, opened 2026-09-19
fix(drift): treat qualified-name symbol notation as not a path

dyk1454683243-sudo:
## What

Treat symbol / qualified-name notation such as `` `src/auth/login.validateToken` `` as **not a filesystem path**, so it never becomes a path claim or `MISSING_PATH`.

`isNotAPath` now rejects values whose final slash-segment matches `identifier.identifier` (JS/TS method or property) and does **not** end in a known file extension. Real files stay claims: `src/auth/login.ts`, `.mex/ROUTER.md`, `package.json`. Extra dotted segments such as `foo.d.ts` are left for #216.

## Why

**Addresses #202 (bullet 1 only).**

Do **not** treat this as `Fixes #202` or `Closes #202` — that would auto-close the whole mega-issue.

Remaining #202 bullets stay open:

2. Compound extensions (`.d.ts`) — owned by #216
3. Hypothetical examples (`./x`, `x.ts`, `x/index.ts`)
4. Alternation shorthand (`tests/grounding|traversal|…`)
5. Scoped-package resolve (`@scope/pkg/package.json` vs bare specifier) — owned by #217
6. gitignore trailing-slash directory exemption — owned by #218

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/Tooling

## How to test

1. `npx vitest run src/drift`
2. Confirm markdown that cites `` `src/auth/login.validateToken` `` as a `qualified_name` example produces **zero** path claims and **zero** `MISSING_PATH`.
3. Confirm a real missing file such as `src/auth/login.ts` is still reported as `MISSING_PATH`.
4. Confirm existing #143 cases stay green, and `.mex/ROUTER.md` / `package.json` still extract as path claims.
5. `npm run typecheck`

## Checklist

- [x] Focused drift tests pass (`npx vitest run src/drift` — 12 passed, 1 skipped `#201` version-shaped case left untouched)
- [x] Claims extraction tests pass (`npx vitest run test/claims.test.ts` — 30 passed, including the bullet-1 skip plus preserved real paths)
- [x] Typecheck passes (`npm run typecheck`)
- [ ] Tests pass (`npm test`) — full suite is not the gate for this `isNotAPath` filter
- [x] No breaking changes (or documented below)
- [ ] Tested locally with a real project

## Code-graph changes

- [ ] This PR targets `main`
- [ ] A linked issue agrees on the bounded extractor/resolver scope
[...]

PR #221, opened 2026-09-19
fix(drift): treat pipe-alternation shorthand as not a path

dyk1454683243-sudo:
## What

Treat pipe-alternation shorthand such as `` `tests/grounding|traversal|edges|store.test.ts` `` as **not a filesystem path**, so it never becomes a path claim or `MISSING_PATH`.

`isNotAPath` already dropped globs with `*`/`?` but not `|`. A token with `|` separating path segments or basenames is prose shorthand, not a file. Real paths without pipes stay claims: a missing `tests/store.test.ts` is still `MISSING_PATH`.

No fancy parser — the issue token is a single inline-code string containing `|`.

## Why

**Addresses #202 (bullet 4 only).**

Do **not** treat this as `Fixes #202` or `Closes #202` — that would auto-close the whole mega-issue.

Remaining #202 bullets stay open:

1. Symbol notation (`src/auth/login.validateToken`) — owned by #220
2. Compound extensions (`.d.ts`) — owned by #216
3. Hypothetical examples (`./x`, `x.ts`, `x/index.ts`) — owned by #222
5. Scoped-package resolve (`@scope/pkg/package.json` vs bare specifier) — owned by #217
6. gitignore trailing-slash directory exemption — owned by #218

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Refactor
- [ ] Docs
- [ ] CI/Tooling

## How to test

1. `npx vitest run src/drift`
2. Confirm markdown that cites `` `tests/grounding|traversal|edges|store.test.ts` `` produces **zero** path claims and **zero** `MISSING_PATH` for that token.
3. Confirm a real missing file such as `tests/store.test.ts` is still reported as `MISSING_PATH`.
4. Confirm existing #143 cases stay green.
5. `npm run typecheck`

## Checklist

- [x] Focused drift tests pass (`npx vitest run src/drift` — 11 passed, 1 skipped `#201` version-shaped case left untouched)
- [ ] Tests pass (`npm test`) — full suite is not the gate for this one-filter change

## Code-graph changes

- [ ] This PR targets `main`

This is a drift claim-extraction filter only. It does not change graph identity, extractors, or the other #202 bullets. Issue #202 stays **open**.


dyk1454683243-sudo:

